PR #1146, opened 2014-12-29
Test key events fix3

dchambers:


thecapdan:
@dchambers while the tests are passing, for some reason each library is logging the bundles it's creating, e.g.:

```
Testing libs/javascript/br-i18n/test-unit (UTs):
bundles/i18n/i18n.bundle
bundles/js/js.bundle
......................................................................
.....
Total 75 tests (Passed: 75; Fails: 0; Errors: 0) (34.00 ms)
  PhantomJS  : Run 75 tests (Passed: 75; Fails: 0; Errors 0) (34.00 ms)
Tests Passed.
```

Can be seen here:
https://travis-ci.org/BladeRunnerJS/brjs/builds/45363011

This doesn't seem to be happening in develop so I assume this is an issue specific to this PR(?)

thecapdan:
The issue mentioned above is resolved.

thecapdan:
Tests are failing against the version of chromium and firefox that we use in our workspace. Throws this exception:

```
error loading file: /test/bundles/js/js.bundle:17895: keyboard-event polyfill unable to shim browser.
```

dchambers:
This is only happening because we are testing with really old versions of Firefox and Chrome. We should finally take the opportunity to update to recent versions of Firefox and Chrome on all operating systems.

dchambers:
@andyberry88, I changed 'dependencies.gradle' to use the same version of Firefox and Chrome as used by CT, but I guess that doesn't work without extra work to make the browsers available outside of Caplin?

andyberry88:
@dchambers The new browser versions also need to be uploaded (checked in) to the build dependencies repo/site at https://github.com/BladeRunnerJS/brjs-build-dependencies. Theres a script that will help with the upload to make sure its in the right place and all the Maven metadata is updated.

thecapdan:
@andyberry88 we have stumbled across a problem with our test-runner.conf. 'firefox' and 'firefox-webdriver' are pointing at the same version of firefox.

the version of firefox that is put in our build/browsers directory depends on which of these lines runs second in build.gradle:

```
from { configurations.browsers_dep.collect { browser -> zipTree(browser) } }
from { configurations.browsers_webdriver.collect { browser -> zipTree(browser) } }
```

we need to ensure they are not sharing a folder name - that way we can distinguish between them when we refer to them in test-runner.conf. will speak to you about it on wednesday

thecapdan:
for now I have reversed the lines above since it's more important for us to have an up to date version of firefox. BUT, even when we use a newer version of firefox, i still get test failures. here's a snippet of the stack trace:

```
Total 19 tests (Passed: 15; Fails: 4; Errors: 0) (209.00 ms)
  Firefox 29.0 Windows: Run 19 tests (Passed: 15; Fails: 4; Errors 0) (209.00 ms)
    User performs text events. simulates a user typing 1234 into a textbox
         failed (10.00 ms): TypeError: TypeError: Illegal constructor. in http://localhost:4224/test/bundles/js/js.bundle (line 20551)
            DL4KeyboardEvent@bundles/js/js.bundle:20551
            Utils.fireKeyEvent@bundles/js/js.bundle:35361
            TypedValue.prototype.set@bundles/js/js.bundle:36062
```

dchambers:
@thecapdan, this is failing because Firefox 29 claims to support DOM Level 4 Keyboard Events, but actually doesn't support it properly. This was fixed in Firefox 31 (released in July 2014). I suggest we just upgrade to an even more recent version, which I'll take a look at doing tomorrow morning.

thecapdan:
great - now have it passing in IE, chrome and an up to date firefox. merging

Issue #1152, opened 2015-01-07
NumericValidator does not consider floats or negative numbers as valid


avsmithy:
Floats and negative numbers should be considered valid according to `br/presenter/validator/NumericValidator`

```
m_oRegex = new RegExp(/^\d+$/);
m_oRegex.test(0.1) == false
m_oRegex.test(-1) == false
```

dchambers:
@avsmithy, don't suppose you happen to have a pull-request for us? ;-)

PR #1156, opened 2015-01-07
NumericValidator validates floats and negatives

avsmithy:
fixes https://github.com/BladeRunnerJS/brjs/issues/1152


dchambers:
Dev reviewed :+1:

avsmithy:
The strings `".1"` and `"1."` will fail this regex. If parsed into a number before validating, they will be converted to `0.1` and `1` respectively and therefore pass.

dchambers:
Hi @avsmithy, I just took a look at the code and it's both. The `vValue` parameter has a _variant_ type, hence the `v` prefix. Makes things more painful doesn't it!

avsmithy:
@dchambers Thanks for the info, I've made some changes if you'd like to review them.

PR #1159, opened 2015-01-08
Import blade from default bladeset 1133

ioanalianabalas:

### DIFF
--- a/brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
+++ b/brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
@@ -4,8 +4,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


dchambers:
Hi @ioanalianabalas, just being code reviewing, and I actually found it hard to understand your changes, even though I helped you make them!

For starters it isn't obvious to me what "Changed expected require path to remove default; removed the use of hashmaps." actually means.

Secondly, things have been made more difficult due to the fact that the commit contains green-to-green preparatory re-factoring plus actual code changes all in the same commit. In the same way that these changes should be made in distinct phases (see Martin Fowler's [workflows of refactoring](http://martinfowler.com/articles/workflowsOfRefactoring) for example), I think it makes sense that the commits are kept apart too, to simplify comprehension.

Finally, the final two changes just seem to be amendments to the earlier commit, whereas if these had been committed using `git commit --amend` this would have made things simpler again. I'll probably wait to review this until you're back in the office and you can walk me through it.

Issue #747, opened 2014-06-19
move all BRJS plugins into a separate project(s)


andyberry88:
We have several instances in the model where the behaviour of a plugin is relied on to make the model work (e.g resources). We should move all plugins into a separate project that depends on core and use mock plugins to test the model behaviour. Because brjs-core does not depend on the plugin project the separation of classpaths/dependencies will prevent the model depending on plugin features.

PR #1160, opened 2015-01-08
Seperate plugins into separate projects

andyberry88:
PR for #747 

- starting to move plugins into their own separate projects
- temporarily moving all brjs-core tests into a separate project so we can slowly move plugin specific tests into plugin projects and core tests back into brjs-core
- removing the need for `workspace` and `build-resources` directories in `brjs-sdk`
- fixing Gradle config so it doesnt run tasks that dont need to be run

**This is not 100% complete but it would be good to get it merged in since we're otherwise going to forever have merge conflicts**


dchambers:
Dev reviewed :+1:

thecapdan:
No matter what I try - I can't get this branch to work. It seems to assemble correctly - but when i try to run any brjs command i get the following message:

```
```

I check the libs directory and there is no java folder (not sure if this is by design!). Moving back to dev

andyberry88:
@thecapdan The problem above was caused by the Gradle config that now auto imports all sub projects only working on Linux due to the way it uses file paths to calculate the project name. It should now be fixed.

thecapdan:
working now on windows. builds successfully, tests pass. merging

PR #1164, opened 2015-01-08
Usage stats plugin

andyberry88:
- adding the anonymous stats usage plugin
- adding some new events in the model
- command runner now asks if the user will allow stats tracking and provides `--stats` and `--no-stats` flags to immediately enable/disable stats which the build also uses 

this should be merged in after #1160 since its taken off of that branch.


dchambers:
@andyberry88, was getting rid of the 'workspace' directory a pre-requisite for the usage plug-in, or just something you decided to do at the same time?

andyberry88:
Getting rid of the 'workspace' directory was done as part of the separating plugin projects work. This branch was taken based on #1160 which hasn't been merged yet so those changes are showing up in this PR too.

dchambers:
Ah, okay, so it was only the penultimate three commits that I needed to review!

Everything seems okay except that the description seems inaccurate since this feature has not been made available as a plug-in as described, but has instead been made available as part of the core of BRJS. Provided this is acceptable to everybody then this issue can be considered to be dev reviewed.

thanhc:
Just got a few questions regarding the events we fire:
- Turning the stats on and off using the --stats and --no-stats causes an install event to be fired each time you turn it on, it very unlikely people will be changing this.
- loading the dashboard causes the aspect event to fire twice - should we be monitoring the dashboard app?
- running commands via the dashboard does not fire any command events such as create-app
- running test causes lots of defaultTestPack events to be fired, I ran two but 8 events triggered.

andyberry88:
`Turning the stats on and off using the --stats and --no-stats causes an install event to be fired each time you turn it on, it very unlikely people will be changing this.`
 - Without storing some persistent data we can't know whether we've already fired a new install event for this machine. As you say this is probably a relatively minor issue which is unlikely to happen. 

`loading the dashboard causes the aspect event to fire twice - should we be monitoring the dashboard app?`
 - I've added a check so we ignore any system applications.

`running commands via the dashboard does not fire any command events such as create-app
 - This has been fixed. Since the dashboard uses BRJS to locate commands by name we can also trigger the event when dashboard commands are used.

`running test causes lots of defaultTestPack events to be fired, I ran two but 8 events triggered.`
 - Since an event is triggered every time we create a BundleSet multiple events will be fired for a single test suite run.


Since the above have been fixed moving this back into ready for test.

thanhc:
I tested the stats plugin and most of it now all works, we are recording some of the dashboard commands such as create-app/bladeset/blades but not export-app/jsdoc 
@andyberry88 are you happy for this to be a new issue and mark this ticket as done or move this one back to dev?

andyberry88:
I've pushed a fix to develop (https://github.com/BladeRunnerJS/brjs/commit/9ea5767fe00d41885fb9ecb5884f226c164946a8) so all commands now trigger the event and any dashboard commands are prefixed with "dashboard:" so they can be distinguished from CLI commands.

PR #1165, opened 2015-01-09
Add our _assertEquals that checks if it's an array and calls assertEqual...

### LINKED ISSUE #1066: assertEquals uses double equals, this means some assertions will incorrectly pass
In PresentationModelFixture.doThen, when checking a caplin.presenter.property.Property value, assertEquals is used.
Since assertEquals uses double equals, this means some assertions will incorrectly pass.”

Our case for this is that we want to check for a string type numeric value formatted with trailing zeroes, e.g. '1.4000', but the property's formatted value could be 1.4, '1.4' or '1.4000' and the assertion would still pass (meaning we haven't verified that the value is correctly formatted, only that the value is numerically equivalent).

We've changed this to use assertSame, which did cause a number of tests to fail, however the fixes are simply to ensure that number type values aren't incorrectly wrapped in single quotes, and string type numeric values are all correctly wrapped in single quotes. This was something that appeared to be arbitrary in the past - now we understand why.

It's not hugely important on our side that you also make this change, but we felt that it improved the tests, so we wanted to raise it with you. It could also warrant investigation of some of the other assertions that are made (on non-property items).

https://jira.caplin.com/browse/PCTLIBRARY-1515

ioanalianabalas:
...s if it is, and assertSame if it is not.

Fixes #1066.

### DIFF
--- a/brjs-sdk/workspace/sdk/libs/javascript/br-presenter/src/br/presenter/testing/PresentationModelFixture.js
+++ b/brjs-sdk/workspace/sdk/libs/javascript/br-presenter/src/br/presenter/testing/PresentationModelFixture.js
@@ -92,6 +92,15 @@ br.presenter.testing.PresentationModelFixture.prototype.doWhen = function(sPrope
 	this._doGivenAndDoWhen(sProperty, vValue);
 };
 
+function _assertEquals(msg, expected, actual) {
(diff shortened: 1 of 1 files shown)


dchambers:
Dev reviewed :+1:

Issue #1059, opened 2014-11-10
base tag breaks links in single page apps


leggetter:
Since we have a base tag which goes up on directory any relative links for routes in single page apps actually reload the page because the link goes to a different page.

![screen shot 2014-11-10 at 19 02 02](https://cloud.githubusercontent.com/assets/328367/4981770/5429660c-690c-11e4-90c7-c75295ca54bd.png)

If this is something we have to put up with, what is the solution? Hard-coding the locale into the `href` doesn't feel right.

dchambers:
This should be fixed by #1173.

thecapdan:
We are looking at an issue with this PR that causes an app to be redirected to the login page when you try to login. Will move back to dev for now....

thanhc:
This has been tested and verified by adding an anchor tag to index.html and ensuring that page is not reloaded, tested with both single and multi locale apps

Issue #692, opened 2014-06-10
Remove the /<locale> from the end of URLs if there is only a single locale


andyberry88:
If an app only has a single URL /, rather than /en, should be used as the app home page.

As part of this the logic to do with the request path for bundles should be moved to a centralised place and the '../' is only added to bundle URLs if the content is being served from /app/<locale>.

minlokchan:
More information can be found here:  http://yourls.caplin.com/9

dchambers:
I've now investigated whether we do still really need to use a `<base/>` tag anymore, and proved that we don't (see the [deprecate-base-tag](https://github.com/BladeRunnerJS/brjs/tree/deprecate-base-tag) branch for example). Instead, it turns out there is a compromise to be made:

  1. If we don't have a base tag then `unbundled-resources` URLs break.
  2. If we do have a base tag then [route URLs break](https://github.com/BladeRunnerJS/brjs/issues/1059).

It happens that not requiring single-locale apps to include the locale in their URL will reduce the likelihood that developers encounter either of these problems, but in some ways will make it worse since library code and shared templates that need to work in both single-locale and multi-locale apps will need a dynamic solution.

My vote is to break `unbundled-resources` URLs by default for the following reasons:

  1. The `unbundled-resources` content plug-in is the only content plug-in that end-developers need to be aware of, but for which we don't provide a tag to simplify access to it's resources.
  1. `unbundled-resources` URLs are usually found within HTML only, whereas route URLs are also commonly found within both HTML and JavaScript, increasing the chance that end-developers need to use a programmatic solution, in addition to a simple tag.

@james-shaw-turner / @andyberry88 / @thecapdan: what do you think?

stevesouth:
I have had issues this week with caching in chrome. I still believe
versioning is useful in dev mode.

dchambers:
Hi @stevesouth, we're not talking about getting rid of versioning here, we're talking about both removing the need for the locale to appear in the URL of single locale apps, and removing the need for a `<base/>` tag.

In BladeRunner we used to use a `<base/>` tag to allow us to have versioned URLs, but in BladeRunnerJS we've solved that differently, though we still ended up keeping a `<base/>` tag for other reasons, and that was probably a dubious decision, which I'm arguing against here.

dchambers:
Just noticed the 'index.html' page within the dashboard:

```
<@js.bundle dev-minifier="combined"@/>
[...]

PR #1173, opened 2015-01-14
Deprecate base tag and don't display locale in URL for single-locale apps

dchambers:


dchambers:
This is probably going to break until [snapshots.repository.codehaus.org](http://snapshots.repository.codehaus.org/) starts working again, but can be dev-reviewed now anyway.

dchambers:
Replicating took a bit of effort. Here are some notes in case I need to do this again. I started by running these commands:

```bash
./brjs create-app app1
./brjs j2eeify app1
./brjs serve
```

after which I confirmed the app worked before enabling authentication by adding this to the `web.xml` of my new app:

```xml
<session-config>
	<session-timeout>10000</session-timeout>
</session-config>

<security-role>
	<role-name>user</role-name>
</security-role>

<security-constraint>
	<web-resource-collection>
		<web-resource-name>Application</web-resource-name>
		<url-pattern>/*</url-pattern>
	</web-resource-collection>
	<auth-constraint>
		<role-name>user</role-name>
	</auth-constraint>
</security-constraint>

<security-constraint>
	<web-resource-collection>
		<web-resource-name>Login Section</web-resource-name>
		<url-pattern>/login/*</url-pattern>
	</web-resource-collection>
</security-constraint>
```

after which I refreshed the browser and confirmed that a username was required, and then updated `app.conf` to add a second locale, and confirmed the app still worked, but contained the locale within the URL.

By adding this to `web.xml`:

```xml
<login-config>
	<auth-method>FORM</auth-method>
	<realm-name>BladeRunnerLoginRealm</realm-name>
	<form-login-config>
		<!-- TODO: see if it's possible to have the same config for both Tomcat and Jetty -->
		<!-- Tomcat does not 302 redirect to /APP_CONTEXT/login as search engines don't like 302's but Jetty does hence the two form-login-pages -->
		<!-- start-env: dev -->
		<form-login-page>/login/</form-login-page>
		<!-- end-env -->
		<!-- start-env: prod
		<form-login-page>/login/unbundled-resources/redirect.html</form-login-page>
		end-env -->
		<form-error-page>/login/</form-error-page>
	</form-login-config>
	<auth-constraint>
		<role-name>*</role-name>
	</auth-constraint>
</login-config>
```

and restarting the server I was able to observe the browser forward to the `/app1/login` page, but this gave me a 404 since new apps don't have login aspects.

By copying the entire `login-aspect` from 'fxtrader' to my new app, I was finally able to replicate.

dchambers:
Okay, this seems to be an issue inside CTs login aspect. By replacing the login aspect with a single `index.html` file containing the following content, I was able to login just fine:

```html
<!DOCTYPE html>
<form class="formLogin" id="formLogin" action="j_security_check">
	<input name="isLoginPage" type="hidden" value="login-page" />
	<input id="txtUsername" class='username' type="text" name="j_username" size="15"
	 value="user1@caplin.com" autocorrect="off" autocapitalize="off"/>
	<input id="txtPassword" class='password' type="password" name="j_password" size="15"
	 value="password" autocorrect="off" autocapitalize="off"/>
	<button type="submit">Login</button>
</form>
```

I guess this will be caused by our changes to the base tag, or the need to now prefix requests for `unbundled-resources` with `../` (e.g. via the `<@unbundled-resources@/>` tag).

thecapdan:
Thanks @dchambers - I'll retest and see if i can work out how we can word this in our release notes if it's going to effect users outside of CT

dchambers:
Hi @thanhc, I can now see where the confusion has come from regarding this issue and #1180. To clarify, there are 3 distinct locale related bits of work that have been done:

  1. Not displaying the locale for single-locale apps (this _pull-request_).
  2. Deprecating the `<@base.tag@/>` tag (this _pull-request_).
  3. Making locale handling more flexible (#1180).

Unfortunately, 1 and 2 had to be done together as they are affected by each other. I've now renamed the title of this _pull-request_ to make that clearer.

thanhc:
After testing this issue, here are my findings:
- In the brjs template the index.html still has includes the <@base.tag@/> (if we are deprecating this then we should remove it from our templates)
- upgrading CT to use this version of brjs is not so easy, as its not just a case of find and replace, there are a few request to unbundled-resources and servlets which all need the ../ added to the front of them for them to work in a multi-locale app, in addition to this there are some places where there is a function which adds a part of the url in such as the popouts: var ticketPopout = new caplin.popout.Popout('popout/'); which requires adding ../ in front of it.
- Also we needed to make a change to a webcentric js library file to also include the ../ as this makes a request to servlet/webcentric/logins/

dchambers:
We can get around the backwards compatibility issue by allowing `enableDeprecatedBaseTag: true` to be added to `app.conf`, in which case we will:

  1. Start emitting a `<base href="../"/>` tag again (alongside the deprecation message) when `<@base.tag@/>` is used.
  1. Stop prefixing versioned bundle paths with `../`.

@andyberry88 / @james-shaw-turner: are you happy with this?

andyberry88:
Do we need to get around the backwards compatibility issue? It's a relatively small change that's going to have to be made at some point if we are deprecating the feature and removing it at some point. I can see a `enableDeprecatedBaseTag` sticking around forever and just becoming bloat we don't need only to save a small amount of time.
My preference is to spend the time investing in docs explaining why we have `../` in the paths and how developers can use the `undbundled-resources` tag and `AppMeta` service to automatically prepend that if the app is using more than 1 locale.

adamshone:
@andyberry88 Yes, I think we do have to think about the backwards compatibility issue. Upgrading to BRJS introduced the "/en" to the application URL, which meant that our clients were required to find and pre-pend an extra "../" to every relative path in their applications. They will justifiably be annoyed if we put them through this process to upgrade to CT 4.0.0, and then make them change it all back again when they upgrade to CT 4.0.2. It makes it look like we don't know what we're doing and gives a bad first impression of BRJS.

dchambers:
@adamshone, yes, we do need to properly evaluate all the options so as to cause as little upgrade pain as possible. Although the need for these changes was known about for quite some time, CT releasing on a pre 1.0 release of BRJS has just made our ability to complete these stories that much more difficult.

andyberry88:
`Upgrading to BRJS introduced the "/en" to the application URL, which meant that our clients were required to find and pre-pend an extra "../" to every relative path in their applications.`

We should get people to migrate to using the meta service anyway. If single locale apps aren't going to have the `/en` and multi locale apps will then the `../` won't always be required. Do we know how many people are going to have to make these changes or whether most of it is contained within the motif codebase?

adamshone:
They tend to all do logout a bit differently, but apart from that most of the changes are in the motif. I had another look at it and actually the references to "unbundled-resources" don't seem to have "../" in most cases. I think we must have removed that in favour of the @base-tag@. 

I had a chat with Dom and he reckons that if we change it so that the locale is never in the URL, even in a multi-locale app, then the upgrade should be painless. We just deprecate the @base-tag@ and change it to be inert, and then all of the existing "unbundled-resources" strings should still resolve correctly.

dchambers:
There are lots of subtleties to this issue:

  1. The `<base/>` tag makes routed SPA apps [really awkward](https://github.com/BladeRunnerJS/brjs/issues/1059) to create.
  1. Having a different URL structure for _single-locale_ and _multi-locale_ apps makes unversioned requests [really awkward](https://github.com/BladeRunnerJS/brjs/issues/692#issuecomment-69958254) to create &mdash; at present we don't even offer a tag or API for making unversioned requests.
  1. Versioned `unbundled-resources` URLs that occur outside of the index page are awkward, but these are necessary for static file deployment.
  1. We have to support unversioned `unbundled-resources` URLs since these sometimes find their way into databases, and since the App-Cache plug-in uses a manifest instead of versioned URLs.
  1. Making unversioned `unbundled-resources` URLs easier to work with than versioned URLs is bad, since it results in slower app start-up times.

Therefore I propose that we make the following changes:

  1. We change multi-locale apps to no longer contain the locale in the URL, and to dynamically load the resources instead.
  1. We ensure versioned URLs are used wherever possible, and add comments explaining why unversioned URLs have been used when they were genuinely needed.

dchambers:
This pull-request is superseded by #1206, which is based off of it.

PR #1176, opened 2015-01-16
Updating Selenium so we can use the latest version of Firefox.

dchambers:


thecapdan:
tested with ITs on the brjs todo app + compatible with v35 of firefox and older versions too

PR #1180, opened 2015-01-19
More flexible locale handling

dchambers:
The user is switched to the _active-locale_ using the following code:

```js
require('br-locale-switcher').switch();
```

wherewithin the _active-locale_ is determined using this snippet of code:

```js
require('br-locale-utility').getActiveLocale(appLocales, localeCookieName);
```

You can change how the _active-locale_ is calculated by overriding the `br-locale-utility` library within your app, and you can change how _locale-switching_ is implemented by overriding the `br-locale-switcher` library within your app.


dchambers:
There's still more work to do as there's an unhealthy relationship between the _locale-service_ and `br-locale-utility`, and since we need to update the main site documentation too, but I'm parking this work for the moment while I circle back and deal with #1173.

dchambers:
The team have decided to merge what's been done so far now since it's a large improvement to what we have now, then improving `br-locale-utility` and the _locale-service_ in a separate _pull-request_.

dchambers:
@andyberry88, can you dev-review this please?

brjs-core/src/main/java/org/bladerunnerjs/model/AbstractJsLib.java
```diff
@@ -42,6 +42,18 @@ public AbstractJsLib(RootNode rootNode, Node parent, MemoizedFile dir)
 	}
 	
 	@Override
+	public List<LinkedAsset> modelSeedAssets()
```
andyberry88: `modelSeedAssets` doesn't feel right to me, we also don't talk about 'model' seeds in other methods. `seedAssets` may be better here.

andyberry88:
@dchambers Only one comment on this (see above).

brjs-core/src/main/java/org/bladerunnerjs/model/AbstractJsLib.java
dchambers: Fair enough, but `modelSeedAssets()` is a pre-existing abstract method on `AbstractBundlableNode`, and is nothing to do with this commit per-se. Maybe add a comment to the plug-in review issue?

andyberry88: Yep OK. I'll change it as part of my plugin API changes.

dchambers:
I've decided to close this pull-request as I believe merging it will cause merge conflicts in the future, once #1206 is merged.

PR #1188, opened 2015-01-22
Require support index pages 2 720

### LINKED ISSUE #720: Index Pages Should Have Proper require() Support
At the moment index pages are using a trie to find dependencies, but it would be preferable to also use the same `require()` scanning we use for CommonJs classes. This will enable better feedback to developers that attempt to use an incorrect _require-path_.

ioanalianabalas:
Fixes #720.


dchambers:
Hi @ioanalianabalas, the code here is all really good, but I think the commits are a little confusing to follow because you've mixed preparatory re-factoring (making `addRequirePathsFromReader()` available from a utility) with the main change (adding proper require support for index pages), and then compounded this problem by describing the entire commit as:

> Renamed PrimaryRequirePathUtility to RequirePathUtility;
Moved the CommonJs addRequirePathsFromReader function to RequirePathUtility;
Modified tests that had NamespacedJs require paths.

which fails to mention the main point of the commit in the first place! Anyway, just something to consider in the future...

PR #1189, opened 2015-01-23
Fixtures commonjs 1178

### LINKED ISSUE #1178: The fixtures() function isn't compatible with CommonJS classes
The `fixtures()` function currently allows you to specify a `FixtureFactory` class using dotted notation, where the class is available within the global namespace, for example:

```js
fixtures('pkg.MyFixtureFactory');
```

but doesn't allow you use _require-paths_, for cases where the class needs to be required, for example:

```js
fixtures('pkg/MyFixtureFactory');
```

If the class has neither dots nor slashes, then both approaches should be tried.

ioanalianabalas:
Fixes #1178.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-core/src/br/core/Utility.js
+++ b/brjs-sdk/sdk/libs/javascript/br-core/src/br/core/Utility.js
@@ -35,9 +35,11 @@ function locate(path, root) {
 	}
 
 	else {
-		if (parsePath(path, ".") == undefined) {
-			return parsePath(path, root, "/");
+		var parsedPath = parsePath(path, root, ".");
+		if (parsedPath == undefined) {
+			parsedPath = parsePath(path, root, "/");
 		}
+		return parsedPath;
 	}
 
 };
--- a/brjs-sdk/sdk/libs/javascript/br-core/src/br/core/Utility.js
+++ b/brjs-sdk/sdk/libs/javascript/br-core/src/br/core/Utility.js
@@ -16,7 +16,7 @@ function parsePath(path, root, separator) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ioanalianabalas:
@dchambers, thank you for your feedback. I have now removed the space between the if and else blocks.

andyberry88:
`
+  if (path.indexOf(".") > -1) {
+		return parsePath(path, root, ".");
+	}
+
+	if (path.indexOf("/") > -1) {
+		return parsePath(path, root, "/");
+	}
+	else {
+		var parsedPath = parsePath(path, root, ".");
+		if (parsedPath == undefined) {
+			parsedPath = parsePath(path, root, "/");
+		}
+		return parsedPath;
+	}
`

@ioanalianabalas I'd suggest adding an extra guard around the first 2 `if` statements, e.g. `if (path.indexOf(".") > -1 && path.indexOf("/") == -1)` etc. Given `a.b/c` it would currently parse it as if it had only `.` separators, the same goes for `/`. There should probably be tests for these scenarios too.
As @dchambers says above, you could also use `if -> else if -> else` which will make it more readable.

ioanalianabalas:
@andyberry88, thanks for your feedback. I have now made these updates.

avsmithy:
@ioanalianabalas can you check the I've made in https://github.com/BladeRunnerJS/brjs/commit/bd23109621e92edea780b9fdfddfe75d95e7a789?

Does the JSDoc for the locate function need to be updated?

ioanalianabalas:
@avsmithy, the changes you've made look reasonable to me. :+1: 

And yes, I forgot to update the JSDoc. Will do that now.

ioanalianabalas:
@avsmithy, I have now modified the JsDoc to reflect the new functionality.

PR #1193, opened 2015-01-27
Changed the redirection path if URL is from Apache Cordova.

### LINKED ISSUE #923: Locale redirection doesn't work in built app when deployed using Apache Cordova
@soruban has found that, when using Apache Cordova, 'index.html' redirects to the wrong URL in the exported app because the `host` variable within that page contains an unneeded slash. I've asked @jngo00 to provide details about the value of `host` when using Apache Cordova.

ioanalianabalas:
Fixes #923.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-locale-utility/LocaleUtility.js
+++ b/brjs-sdk/sdk/libs/javascript/br-locale-utility/LocaleUtility.js
@@ -117,10 +117,18 @@ LocaleUtility.getLocalizedPageUrl = function(pageUrl, locale) {
 	var protocol = urlParser.protocol;
 	var host = urlParser.host;
 	var url = urlParser.pathname;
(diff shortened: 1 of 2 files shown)


dchambers:
I don't like this _pull-request_ as `LocaleUtility.js` shouldn't have any code in it to do with Apache Cordova. Can this not be fixed in a generic way?

ioanalianabalas:
@dchambers, thank you for your feedback. I have now changed the name of the variable as it is not Cordova-specific.

avsmithy:
@ioanalianabalas I've modified the test case, it should work on Cordova once this passes.

dchambers:
I think in general the use of comments is sign that the code is unreadable, and could be made easier to read.

avsmithy:
@dchambers `.endsWith` is ES6. I've tried to improve the readability of the method.

dchambers:
@avsmithy, yes, the method is now infinitely more readable than it was before :+1:

avsmithy:
Static file and Cordova builds are now working.

PR #1202, opened 2015-02-06
explicitly cleaning up temporary directories

andyberry88:
explicitly cleaning up temporary directories rather than relying on the
shutdown hook to do it since it seems JVM shutdown hooks sometimes arent
triggered when running via Eclipse


thanhc:
tested and verified, ran a number of test from eclipse and verifying temp data is being removed

PR #1205, opened 2015-02-09
Modified brjs so that paths are converted to Cygwin format for the java ...

### LINKED ISSUE #1150: in vanilla cygwin "brjs" script does not run
Issue:

waqask@WAQASK-T430 /cygdrive/c/Users/waqask/Desktop/BladeRunnerJS/sdk
$ ./brjs start

The reason the script behaves like this is because Java under cygwin cant map "/cygwin/c/users/somedir" or "/c/users/somedir" to "C:\Users\somedir". 

When you run the "brjs.cmd" script, the cp provided has the actual "C:\", hence that will work.

sJira: http://yourls.caplin.com/7

ioanalianabalas:
Fixes #1150.

### DIFF
--- a/brjs-sdk/sdk/brjs
+++ b/brjs-sdk/sdk/brjs
@@ -21,8 +21,10 @@ fi
 SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
 BRJS_CLASSPATH="$SCRIPT_DIR/libs/java/system/*:$SCRIPT_DIR/../conf/java/*"
 
-# UNIX & GitBash version
-java $JAVA_OPTS -cp "$BRJS_CLASSPATH" org.bladerunnerjs.runner.CommandRunner "$SCRIPT_DIR" "$@"
-
-# Cygwin version (uncomment this line for Cygwin)
-# java $JAVA_OPTS -cp "`cygpath -wp $BRJS_CLASSPATH`" org.bladerunnerjs.runner.CommandRunner "`cygpath -wp $SCRIPT_DIR`" "$@"
+if [ "$(echo $(uname -s) | cut -c 1-6)" = "CYGWIN" ]; then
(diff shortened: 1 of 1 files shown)


avsmithy:
Could we use an if statement to detect Cygwin instead?

```
if [ "$(expr substr $(uname -s) 1 6)" == "CYGWIN" ]; then
	# Cygwin version
	java $JAVA_OPTS -cp "`cygpath -wp $BRJS_CLASSPATH`" org.bladerunnerjs.runner.CommandRunner "`cygpath -wp $SCRIPT_DIR`" "$@"
else
	# UNIX & GitBash version
	java $JAVA_OPTS -cp "$BRJS_CLASSPATH" org.bladerunnerjs.runner.CommandRunner "$SCRIPT_DIR" "$@"
fi
```

avsmithy:
I needed to run dos2unix to fix line endings before it would work in Cygwin. Can we do this and check it in?

avsmithy:
Also, we should link to `#!/bin/bash` rather than `#!/bin/sh`.

andyberry88:
> Also, we should link to #!/bin/bash rather than #!/bin/sh.

Not sure about this one. `/bin/sh` is present in pretty much all *nix systems, `/bin/bash` is not, I imagine it could not be in *nix like environments. So long as the script is `POSIX` compliant it should run anywhere with `#/bin/sh`.

avsmithy:
`/bin/sh` can be a [symlink to the default shell](http://askubuntu.com/a/141932), so you could run into inconsistencies. You're correct though, if the script is `POSIX` compliant it should work fine everywhere.

avsmithy:
The new if statement has been tested in Mac, Cygwin and MinGW.

thanhc:
tested and verified on windows/mac/linux

Issue #1186, opened 2015-01-22
Clean locale-service and br-locale-utility


dchambers:
Clean up the unhealthy relationship between the _locale-service_ and `br-locale-utility`.

dchambers:
We currently have a `LocaleService` interface that looks like this:

```js
LocaleService.setLocaleCookie = function(locale, days);
LocaleService.getLocale = function();
LocaleService.getPageLocale = function();
```

and for which the default implementaion is a thin wrapper around `br-locale-utility`. As it stands, this service has a number of faults:

  1. It assumes that all implementations will want to use cookies.
  2. It's `setActiveLocale()` method doesn't help with reloading the app at the correct URL (see #804).
  3. It isn't used by the _locale-forwarding_ page anyway, which just uses `br-locale-utility` directly.

Instead, I'd like to propose that we make available two distinct services that allow maximum flexibility in how locales are handled in BRJS. These would be:

  1. A _locale-provider_ service that allows the _active-locale_ to be provided or updated.
  2. A _locale-switcher_ service that causes the _active-locale_ to be used.

In this way, there could be alternative implementations offered for either or both parts. The service interfaces would like this:

```js
LocaleSwitcher.switch();
```

and:

```js
LocaleProvider.getActiveLocale();
LocaleProvider.setActiveLocale(locale);
```

dchambers:
@andyberry88 are you happy with this?

andyberry88:
@dchambers :+1: sounds good to me.

dchambers:
I just noticed that `br/services/AppMetaService` also insists that there must be a thing called a _locale-cookie_.

PR #1206, opened 2015-02-11
Locale service tidy up

dchambers:
This pull-request depends on these branches:

  * locale-redirection-cordova-923 (#923)
  * more-flexible-locale-handling (#1180)
  * aliasing-require-plugin (#1207)

so mustn't be merged in until these branches have been separately QAed and merged in.

This pull-request fixes the following issues:

  1. Deprecate base tag and don't display locale in URL for single-locale apps (#1173)
  1. More flexible locale handling (#1180)
  1. Clean locale-service and br-locale-utility (#1186)

while providing these additional improvements:

  1. Multi-locale apps no longer have a trailing slash, removing the need for URLs to be conditionally prefixed with `../`.
  1. The default implementation of `br.locale-switcher` (`BRLocaleForwardingSwitcher`) can now be replaced with an alternate implementation (`BRLocaleLoadingSwitcher`) that doesn't require the locale to be a part of the URL (as [suggested](https://github.com/BladeRunnerJS/brjs/issues/692#issuecomment-69958254) by @lukepage), but with the proviso that _view-source_ no longer provides comprehensible output, and that the browser's debugging tools may not always work as well.

This pull request supersedes #1173 (Deprecate base tag and don't display locale in URL for single-locale apps).


dchambers:
You can test the alternate implementation of `br.locale-switcher` by modifying `sdk/libs/javascript/br/resources/aliasesDefinitions.xml` so that:

```
<alias name="br.locale-switcher"
 defaultClass="br.services.locale.BRLocaleForwardingSwitcher" />
```

is replaced with:

```
<alias name="br.locale-switcher"
 defaultClass="br.services.locale.BRLocaleLoadingSwitcher" />
```

dchambers:
@andyberry88, I believe you dev reviewed 1173, which this pull-request supersedes. Can you therefore please dev-review the additional commits I've since added for this pull-request?

dchambers:
One thing worth noting with these changes is that the URL for multi-locale apps that use the default forwarding _locale-switcher_ have now subtly changed. A URL that used to be `myapp/en/` will now be `myapp/en`, with no trailing slash. This allows us to have consistent relative paths for single-locale and multi-locale apps, but does mean that some static web servers (like Python's built-in web-server) can no longer be used to serve multi-locale apps that use the forwarding _locale-switcher_.

This can be worked around by using a web server that allows the file suffix to be inferred (e.g. _jetty_ or _http-server_), or a web server that can be configured to do this (e.g. _nginx_). Alternatively, the loading locale-switcher can be used instead of the forwarding locale-switcher, as it does not add the locale to the URL.

To test with a static web server that can't infer the file suffix I used:

```bash
python -m SimpleHTTPServer
```

and to test with a static web server that can infer the file suffix I used:

```bash
http-server
```

where `http-server` can be installed as follows:

```bash
npm install http-server -g
```

andyberry88:
@dchambers Looks good to me. You have some failing tests though...

dchambers:
Thanks @andyberry88, I'm fixing those now, but will push straight to dev once they're running.

dchambers:
Here's some justification for why I've introduced a new file extension servlet-filter...

If we didn't have to support JSP index pages then it would be easier not to have a file extension, and configure the default file extension within the server. This can be done as follows:

  * Apache (`text/plain` by default, but can easily be changed by setting `DefaultType text/html`)
  * Nginx (`text/plain` by default, but can easily be changed by setting `default_type text/html;`)
  * http-server (`text/html` by default)
  * J2EE servers (no `Content-Type` by default, which most browsers cope with anyway, but can still be changed by providing a default `Content-Type` filter-servlet)

However, so that we can support JSP index pages, it makes more sense to configure the web server to accept URLs that don't contain the file extension, which can be done as follows:

  * Apache (configured using the `MultiViews` feature from `mod_negotiation`)
  * Nginx (configured using `try_files $uri.html $uri/ =404;`)
  * http-server (the `.html` file extension can be omitted by default)
  * J2EE servers (we will need to provide a file extension hiding filter-servlet)

dchambers:
@andyberry88, can I trouble you to dev-review this issue again? There have been a number of commits since you last reviewed it, starting at f499031.

andyberry88:
:+1: Looks good to me.

thanhc:
CT is using the br.services.locale.BRLocaleService but this has been removed from BRJS, would be good if we created a proxy class which called the new functions

thanhc:
Need to add back the alias too

thanhc:
The server filters don't appear to work in multi-locale apps, if you create a new app and add an extra locale and then try and use a jndi key then it does not get replaced correctly, but works for a single locale app

thanhc:
This is now working with the exception of setting the cookie by using br.services.locale.BRLocaleService, but this has been fixed in #1217 where cookies have been deprecated.

dchambers:
@thanhc has noticed that code like `location.href = '../relative-path'` no longer works as a result of this change. Here is the text that will need to be added to the release notes:

### Compatibility Issues

#### Base Tag Plugin Deprecated

The `<@base.tag@/>` tag is now deprecated. You can continue using it within your web pages without breaking anything, but it now emits a comment rather than an actual `<base/>` tag. This change was made so we could better support single-page-applications that use internal anchors (e.g. `#add-item`) for routing.

For the most part, things should continue to work without the base tag since we've made corresponding changes to the URLs to compensate; _single-locale_ apps no longer include the locale within the URL and _multi-locale_ apps have URLs at the same level as the app (e.g. `/app/en` rather than `/app/en/`).

However, because the `<base/>` tag was never respected when modifying `window.location`, apps that previously prefixed relative paths with `../` when using these methods, for example:

```js
window.location.href = '../relative-path';
window.location.replace('../relative-path');
window.location.assign('../relative-path');
```

will no longer need the `../` prefix.

PR #1207, opened 2015-02-11
Use asset require plug-in rather than the asset content plug-in

dchambers:
This pull-request contains some refactoring work needed by both myself and @andyberry88, but which should be functionally identical to how things were before.


mayanks-caplin:
Tested the branch. When a new bladeset for app is created and when the bladeset is launched - error message is shown.

mayanks-caplin:
Tested and looks fine now.

andyberry88:
@mayanks-caplin Should this now be in either the `Test` or `Done` column?

PR #1216, opened 2015-02-13
Better debug messages

### LINKED ISSUE #1209: output timestamp in `--debug` logs
I'm trying to debug a Caplin support issue relating to performance and the `--debug` logs don't provide any timestamp information so it's difficult to get an idea of any delays  in asset discovery etc. 
We should output the timestamp at the start of the log line if the `--debug` flag is used. Alternately if we don't want to clutter the developer console we could automatically write to a `debug.log` file that contains the timestamp when the debug flag is used. This would have the advantage of developers being able to see the console information as most have to pipe the output to a file.

andyberry88:
- logging if MemoizedValue needs to recalculate a value
- logging if the FileMotificationWatcher detects a change
- adding timestamps to all logs if `--debug` is used (fixes #1209)


mayanks-caplin:
Tested and looking okay on branch.

PR #1217, opened 2015-02-13
Deprecate locale cookies

dchambers:
This pull-request is dependent on #1206, and mustn't be merged in till that has been QAed. In it, we stop using cookies to store locales (we now use  [localstorage](https://developer.mozilla.org/en-US/docs/Web/API/Window.localStorage) instead).

Since existing users will have cookies defining their preferred locale, it's important that these cookies continue to be respected the first time the user uses an upgraded version of the app.


andyberry88:
> we stop using cookies to store locales.

How are we now determining which locale to use? Is it solely based on language settings by default unless developers provide a different implementation for the `br.locale-provider` service?

dchambers:
Hi @andyberry88, I'm going to clarify the description, since all I've actually done is to start using [localstorage](https://developer.mozilla.org/en-US/docs/Web/API/Window.localStorage) instead of cookies (see b0a04d2), which have a much simpler API, and can be used now that we no longer need to inform the server what locale we need.

The [localstorage](https://developer.mozilla.org/en-US/docs/Web/API/Window.localStorage) API is supported from IE8. For backwards compatability, I've continued to use cookies as a fall-back.

thanhc:
verified the following scenario's:
- load app with no cookie or local storage - app uses browsers locale 
- load app with local storage - app uses the locale from local storage
- load app with cookie and no local storage - app uses the cookie locale and writes to local storage too
- load app with both cookie and local storage - app uses local storage for locale

Issue #1112, opened 2014-12-02
Fix `FileModificationWatcherThreadTest ` Travis unreliability


dchambers:
There's a reliability issue when running our `FileModificationWatcherThreadTest` tests on Travis, which results in this strange call-stack:
```
org.bladerunnerjs.memoization.FileModificationWatcherThreadTest > usingTheRealWatchServiceDetectsFileNestedChanges FAILED
    java.lang.NoClassDefFoundError: Could not initialize class sun.nio.fs.LinuxWatchService
        at sun.nio.fs.LinuxFileSystem.newWatchService(LinuxFileSystem.java:47)
        at org.bladerunnerjs.memoization.DefaultWatchKeyService.<init>(DefaultWatchKeyService.java:27)
        at org.bladerunnerjs.memoization.WatchKeyServiceFactory.createWatchService(WatchKeyServiceFactory.java:16)
        at org.bladerunnerjs.memoization.FileModificationWatcherThread.init(FileModificationWatcherThread.java:64)
        at org.bladerunnerjs.memoization.FileModificationWatcherThreadTest.usingTheRealWatchServiceDetectsFileNestedChanges(FileModificationWatcherThreadTest.java:223)
org.bladerunnerjs.memoization.FileModificationWatcherThreadTest > usingTheRealWatchServiceDetectsFileChanges FAILED
    java.lang.NoClassDefFoundError: Could not initialize class sun.nio.fs.LinuxWatchService
        at sun.nio.fs.LinuxFileSystem.newWatchService(LinuxFileSystem.java:47)
        at org.bladerunnerjs.memoization.DefaultWatchKeyService.<init>(DefaultWatchKeyService.java:27)
        at org.bladerunnerjs.memoization.WatchKeyServiceFactory.createWatchService(WatchKeyServiceFactory.java:16)
        at org.bladerunnerjs.memoization.FileModificationWatcherThread.init(FileModificationWatcherThread.java:64)
        at org.bladerunnerjs.memoization.FileModificationWatcherThreadTest.usingTheRealWatchServiceDetectsFileChanges(FileModificationWatcherThreadTest.java:210)
```

[...]

PR #1221, opened 2015-02-17
Fix Utils.fireKeyEvent handling of key parameters

avsmithy:


avsmithy:
Looks like the Travis build failed because of https://github.com/BladeRunnerJS/brjs/issues/1112

PR #1229, opened 2015-02-18
In app locale switching support

### LINKED ISSUE #804: allows apps to push users to the locale forwarder if app locale doesnt match the users choice
Create a mechanism that (optionally) allows apps to push users back to the locale forwarder if the current app locale doesn't match the locale calculated from the locale cookie and browser headers.

This will allow users to refresh the page after changing their locale since after the refresh the locales wont match and the browser will be sent to the locale forwarder. It also means if locale pages are bookmarked the correct locale will ultimately still be used.

dchambers:
This pull-request fixes #804, and is dependent on the [deprecate-locale-cookies](https://github.com/BladeRunnerJS/brjs/pull/1217) pull-request, and should not be merged until that pull-request has been separately QAed and merged.

You can test this issue by creating a new app, and adding `de` to the list of `locales` in `app.conf`, then adding some code like this to your app's `index.html`:

```js
setTimeout(function() {
  require('br-locale/switcher').switchLocale('de');
}, 3000);
```

after which your app should automatically switch to the `de` locale.


thanhc:
This is now working as expected, I'm able to switch locale from within app.

Issue #1154, opened 2015-01-07
allows apps to live separately from the SDK


andyberry88:
Once BRJS can be run from the `PATH` (#1130), allow apps to live separately from the SDK so a single SDK can be used with multiple apps that exist in multiple locations.

ioanalianabalas:
Please see #1231.

andyberry88:
Pushing this back into `dev` as it looks like the `FileModificationWatcherThread` won't pick up changes in the `brjs-apps` directory when it doesn't live next to `sdk`.

```
public FileModificationWatcherThread(BRJS brjs, WatchKeyServiceFactory watchKeyServiceFactory) throws IOException
        {
                this.watchKeyServiceFactory = watchKeyServiceFactory;
                this.fileModificationRegistry = brjs.getFileModificationRegistry();
                directoryToWatch = brjs.dir().toPath();
                this.brjs = brjs;
        }
```

ioanalianabalas:
Fixed by #1296.

andyberry88:
A minor change following the sprint demo:

If I'm running a command inside of an app directory don't require `brjs-apps` to be present.  We need to look at how/if we do this.

Questions:
- What if the create-app command is run inside of an app directory with no `brjs-apps`?
- If serve is run do we just serve up a single app? (I imagine yes)
- Do we only support this if running commands directly inside of an app or anywhere inside the app?

thanhc:
closing this ticket and instead opened #1331 for all remaining issues

PR #1231, opened 2015-02-18
Apps separate from sdk 1154

ioanalianabalas:
For #1154.


andyberry88:
@ioanalianabalas This looks good. It might be worth adding 1 more test to use an entirely different temporary directory rather just the parent file of where BRJS is. You can use the `FileUtils` class to create a temporary directory and use that for the working dir (remember to delete the new temporary dir in your `@After` teardown).

ioanalianabalas:
@andyberry88 I added the test.

dchambers:
@ioanalianabalas, there are merge conflicts you now need to resolve.

dchambers:
@andyberry88, are you happy with the extra test @ioanalianabalas added based on your feedback?

dchambers:
I've dev-reviewed the changes that came after that test and am happy with them.

andyberry88:
Yep. Moved to `ReadyForTest`. Although I'm afraid @ioanalianabalas has some more conflicts to fix.

ioanalianabalas:
@andyberry88 I have fixed the merge conflicts.

thanhc:
seems to have a mixup with what should happen when we have both apss and brjs-apps folders, the warning message and site docs say we use brjs-apps but implementation and also for backwards comp we should use apps

andyberry88:
Both the logging and site docs have been updated

thanhc:
creating an app via the dashboard for the first time is throwing the following error:
Successfully deployed 'ggrtrtg1k' app
Exception in thread "Thread-11" java.lang.NullPointerException
        at org.bladerunnerjs.appserver.AppDeploymentFileWatcher.deployApp(AppDeploymentFileWatcher.java:108)
        at org.bladerunnerjs.appserver.AppDeploymentFileWatcher.checkForNewApps(AppDeploymentFileWatcher.java:83)
        at org.bladerunnerjs.appserver.AppDeploymentFileWatcher.run(AppDeploymentFileWatcher.java:57)

after this creating other apps does not throw the error, but if you restart brjs then the error will be displayed when creating a new app, only seems to happen when brjs is running, using the brjs create-app command does not throw the error.

thanhc:
Also should we prevent people from creating an app called brjs-apps, currently it is possible to create an app with this name and so if you are in this folder and then run create-app it will place the new app in this folder.

andyberry88:
> creating an app via the dashboard for the first time is throwing the following error:

We've probably hardcoded the `apps` dir in the `AppDeploymentFileWatcher`. This should use the BRJS instance to get the apps dir.

> Also should we prevent people from creating an app called brjs-apps

IMO this is a real edge case and you deserve what you get if you do it. Do we also stop people creating folders called `brjs-apps` within an app too?

thanhc:
retested issue and the error no longer occurs when creating app from dashboard when there is no apps folder

PR #1242, opened 2015-02-24
Datepicker control enhancements

charliehw:
This attempts to resolve three problems encountered with the jQuery datepicker control adapter.
 1. Inline datepickers are now supported with the 'inline' control option flag. http://jqueryui.com/datepicker/#inline
 2. Options are merged with defaults instead of being applied after initialisation with defaults.
 3. Datepicker click and mouseup events can be encapsulated with the 'stopPropagation' flag. http://stackoverflow.com/questions/22406505/jquery-ui-datepicker-next-and-prev-buttons-only-have-one-ancestor-e-target-p


charliehw:
I have reverted the stopPropagation option. This can be done on a container element and doesn't need to be in the control.

dchambers:
Nice :+1:

PR #1256, opened 2015-02-27
Optionally parse ambiguous dates to end of month or year

charliehw:
This is a reasonably generic enhancement required for some CT work.

If an ambiguous date (i.e. in a format that does not include the day of the month) is parsed, it will now optionally parse it to the end of the month or year, instead of always parsing it to the start.


charliehw:
This is dependent on [#1246](https://github.com/BladeRunnerJS/brjs/pull/1246) and should not be merged before that.

dchambers:
@charliehw / @bit-shifter?

PR #1262, opened 2015-03-04
Added removeValidator for EditableProperty

mklakakilli:
We need removeValidator functionality for validating a field that is dependant on a value of another input.
When the value of the input changes we want to remove existing validator and add new with a new config.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/property/EditableProperty.js
+++ b/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/property/EditableProperty.js
@@ -115,6 +115,38 @@ br.presenter.property.EditableProperty.prototype.addValidator = function(oValida
 };
 
 /**
+ * Removes {@link module:br/presenter/validator/Validator} from validators array.
+ *
+ * @param {module:br/presenter/validator/Validator} oValidator - validator object class
+ * @param {object} [config] - removes only validator with provided config - optional
+ * @returns {boolean} - true if any validator was removed
+ */
+br.presenter.property.EditableProperty.prototype.removeValidator = function(oValidator, config) {
+	var removed = false;
+
+	if(br.Core.fulfills(oValidator, br.presenter.validator.Validator)) {
+		for (var i = 0; i < this.m_pValidators.length; i++) {
+			if (this.m_pValidators[i].validator == oValidator) {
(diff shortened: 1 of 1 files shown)


andyberry88:
Hi @mklakakilli thanks for the pull request. @dchambers can you dev review this please?

dchambers:
Hi @mklakakilli,

Firstly, thanks for taking the time to submit a patch, and sorry for the slow response time. I've looked at your patch, and wonder if it could be simplified a little? At present there is extra complexity due to the fact that both the validator and the configuration need to match what was originally passed in, and this code looks like it be fragile, and a potential source of bugs.

I can think of a couple of alternatives that might allow us to simplify this code:

  1. Return an identifier when `addValidator()` is invoked that can later be passed to `removeValidator()`.
  2. Create a `SwitchableValidator` class that wraps a normal validator, but which provides `enable()` and `disable()` methods to prevent it from being used when not needed.

Would either of these options be workable for you?

dchambers:
Hi @mklakakilli, any chance you let us know whether options 1 or 2 above would be workable for you?

mklakakilli:
Hi @dchambers 
Both of these options will actually work just fine. I think it's your call which one you want to have in BRJS. We just need to disable/remove existing validator when it's not relevant anymore.

dchambers:
Okay, thanks @mklakakilli, I think I'll probably go for option then.

dchambers:
I've created a [pull-request for option 1](https://github.com/BladeRunnerJS/brjs/pull/1282) as described above. It didn't work out quite as neatly as I'd hoped for though since to maintain backwards compatibility I was unable to return the validator identifier from `addValidator()` and instead ended up having to pass a map into `addValidator()` so that the identifier could be returned (see e798d54).

@bit-shifter / @andyberry88 / @mklakakilli: are people happy with this approach, or should we go with the `SwitchableValidator` proposal instead?

dchambers:
@andyberry88, can you dev-review this when you get a chance please?

andyberry88:
Is it this PR that needs dev reviewing and merging or #1282?

dchambers:
Yes, good point, it's #1282 that needs reviewing.

andyberry88:
I've dev reviewed #1282. Since it also fixes the same issue I'll close this. Thanks for the input @mklakakilli.

PR #1263, opened 2015-03-04
Revert "adding failing test + fix for incorrect namespacing of bladesetl...

mayanks-caplin:
Reverts BladeRunnerJS/brjs#1261


andyberry88:
@mayanks-caplin why does #1261 need to be reverted?

andyberry88:
This doesn't need to be reverted now but another test should probably be added.

PR #1269, opened 2015-03-06
Datepicker control should not update with invalid property

charliehw:
This is **not** required for the recent ct work that uses this control.

The datepicker control should not attempt to update its value with an invalid property. This can cause odd results with some invalid inputs, where the jQuery plugin manages to parse them to nonsensical dates.

e.g. "12345" in yymmdd is represented as December 2048 on the datepicker


dchambers:
@charliehw has assured me that the new test he's added breaks if he reverts his code change; I wan't sure since he sets the date to `invalid`, and I wasn't sure that would ever have caused the date values to change.

PR #1274, opened 2015-03-11
Import apps non corrupted images 1237

### LINKED ISSUE #1237: Import app is causing some images to be corrupted
When importing a newly created app, the br-logo.png in unbundled resources is getting corrupted.
Create a new app, then export and import then load the app and the brjs logo does not load, look in the unbundled resources folder and the image is corrupted.

ioanalianabalas:
Fixes #1237.


thanhc:
images are no longer getting corrupted when importing app

Issue #1270, opened 2015-03-09
Date parsing does not always work with custom input formats


charliehw:
`br.parsing.DateParser` does some quasi-localisation in its `_standardizeDateSeparators` method. However, this prevents dates with "." separators from being parsed when supplying a custom input format.

e.g. it fails to parse "09.03.2015" as "DD.MM.YYYY" because it first converts the input value to "09-03-2015".

    var parser = new br.parsing.DateParser();

    parser.parse('09.03.2015', {
        inputFormats: 'DD.MM.YYYY',
        outputFormat: 'YYYYMMDD'
    }); // returns undefined

One solution would be to merge the custom input formats with the defaults. This would fix the majority of cases. Ideally, the responsibility for default date formats would be delegated to momentjs and the DateParser wouldn't attempt to "standardise" the separators.

dchambers:
@charliehw, might it make sense to have separate `DateParser` and `LocalizedDateParser` classes, where the `LocalizedDateParser` would not allow an input format to be specified, or would only allow input formats like `SHORT` and `LONG`? Or, or is the idea of providing localized date parsing just a bad one?

charliehw:
@dchambers, I think localised date parsing is necessary in many cases. It's also common to have to parse in multiple formats, the use case above with only one input format is just heavily simplified. So more often than not, one would have to add both a DateParser and a LocalizedDateParser to a property (does it make sense to add more than one parser to a property? is this possible?).

I think a LocalizedDateParser wouldn't need to have the input format specified at all. It would attempt to parse using all standard locale-specific date formats (e.g. those defined in momentjs).

charliehw:
It just seems that DateParser is doing the bare minimum

dchambers:
I've just taken a look and Moment.js already seems to support parsing localized date formats. For example, if you execute:

```js
moment('13/12/2000', 'l').format('DD/MM/YYYY');
```

You will get `Invalid date`, but if you first execute:

```js
moment.locale('en-GB');
```

then you will be able to parse `13/12/2000` with no problems.

There are lots more locale specific formatting codes available in addition to `l` &mdash; full details available within the Moment.js [Format](http://momentjs.com/docs/#/displaying/format/) section.

This would seem to suggest that we don't need to bother supporting locale specific parsing at all, and should leave it all to Moment.js.

andyberry88:
@dchambers is this something that has been fixed and should it be in the 0.15.2 release?

[...]

Issue #1228, opened 2015-02-18
Upgrade momentjs to version 2.9


james-s-turner:
This version splits momentjs into two parts
1. The core library (11.6K zipped)
2. Library that includes lots of locales (42K zipped)

In many apps they don't need the full set of supplied locales so lots of uneeded code is downloaded. We should probably supply a minimal locales file for English only, then users can create cutdown versions and put that in their app libs.

We have had a Caplin issue raised that the whole of moment is downloaded just to use the date-formatting functions needed for our br.i18n.Translator function. https://jira.caplin.com/browse/PCTLIBRARY-1223
It would be good to not download the whole of moment just to use our i18n

andyberry88:
This sounds like a quick win. One for 1.0?

charliehw:
There is a discrepancy between the way the date parsing and date formatting is handled within brjs which causes some problems when parsing dates in some locales. The DateParser uses momentjs, and its bundled localisation config to parse dates in most cases. However the DateFormatter uses the Translator, which in turn uses i18n properties for localisation.

A specific issue this causes is when parsing dates in MMM format in DE locale. Parsing fails because:
The formatter formats the month name to "Mär" (from an i18n property)
The parser expects the month name to be "Mär." (from the momentjs locale configuration)

There are probably other localisation issues caused by this discrepancy.

charliehw:
The encoding of some of the localised characters is also broken in the moment.js file we are currently using. e.g. German MÃ¤rz should be März

andyberry88:
I've asked @thanhc to verify whether the encoding issues are specific to the 'thirdparty' bundler or the library itself.

PR #1275, opened 2015-03-11
Localised date parser and formatter

charliehw:
These new classes serve to let Moment.js handle localisation of dates. The shortcomings of the existing DateParser and DateFormatter with regards to localisation (described in part [here](https://github.com/BladeRunnerJS/brjs/issues/1270) and [here](https://github.com/BladeRunnerJS/brjs/issues/1228)) cannot be easily addressed whilst maintaining backwards compatibility.

This PR should perhaps not be merged as is. Moment.js locale configuration should be performed somewhere. i.e.

    require('momentjs').lang(appLocale);

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-formatting/.js-style
+++ b/brjs-sdk/sdk/libs/javascript/br-formatting/.js-style
@@ -1,1 +1,0 @@
-namespaced-js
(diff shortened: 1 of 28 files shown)


dchambers:
Mostly very good. Here's some feedback though:

  1. The `LocalisedDateParser` and `LocalisedDateFormatter` class names are slightly disingenuous since the old `DateParser` and `DateFormatter` also offered localization. Since we need a unique name for these new classes, and since these classes actually work, we should clearly explain that the old classes also attempted to provide localization, but that the new ones actually work.
  2. Don't have `LocalisedDateFormatter` using `LocalisedDateParser.prototype.parse` as a static method, and instead create a `LocalisedDateParserUtil.parse` method that both classes can share.
  3. Store a reference to the locale-service (via `require('br.locale-service').getLocale()`) and provide the user's locale to Moment.js when both parsing and formatting.
  4. Allow the locale to be overridden using config if the developer needs to parse or format using a locale other than the user's default.

Issue #1085, opened 2014-11-19
move all API classes into an org.bladerunnerjs.api package and review the Plugin APIs


andyberry88:
We need to make it obvious which classes are part of our API and ensure they don't change between releases.
As part of this we should make sure the APIs are reasonable before a 1.0 release and review the class/method accessibility.

andyberry88:
#1192 contains the `move all API classes into an org.bladerunnerjs.api package` changes. `review the Plugin APIs` part will done in a separate PR taken from the #1192 branch.

andyberry88:
#1277 contains the `review plugin APIs part of this`.

ccpetercc:
I've spotted an issue on develop which occurs when requiring a file for the first time without restarting BRJS. This causes the require to fail.

To recreate 
- create a brjs App.
- Create a new js file in the default-aspect src directory
- Do not require the file anywhere in the app
- run ./brjs serve
- Navigate to the app's URL - it runs successfully
- Add a require statement to App.js to import the js file created in the second step
- Refresh the app's URL - exceptions will be thrown on the client complaining that no definition for the module is found.

If BRJS is restarted, then the exceptions stop

This works on release v0.15.4

PR #1277, opened 2015-03-12
Review plugin apis 1085

andyberry88:
PR for #1085.

- Refactor the API to match the outcome from the plugin review wiki page - https://github.com/BladeRunnerJS/brjs/wiki/Plugin-Interface-Review
- Make aliasing and services a plugin concern
- Remove AssetLocations
- Changes the ordering mechanism to use a priority - this is up for debate since we've said the better way to do this would be to load the active plugins (and the order) from `brjs.conf`


brjs-core-tests/src/test/java/org/bladerunnerjs/plugin/commands/core/HelpCommandTest.java
```diff
@@ -1,10 +1,10 @@
-package org.bladerunnerjs.spec.command;
+package org.bladerunnerjs.plugin.commands.core;
 
 import org.bladerunnerjs.api.model.exception.command.ArgumentParsingException;
 import org.bladerunnerjs.api.model.exception.command.CommandArgumentsException;
 import org.bladerunnerjs.api.spec.engine.SpecTest;
 import org.bladerunnerjs.plugin.commands.core.HelpCommand;
-import org.bladerunnerjs.testing.utility.MockCommandPlugin;
+import org.bladerunnerjs.testing.utility_copy.MockCommandPlugin;
```
dchambers: Was `utility_copy` meant to be a temporary package name?

brjs-core/src/main/java/org/bladerunnerjs/model/DirectoryAsset.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.Set;
+
+import org.bladerunnerjs.api.Asset;
+import org.bladerunnerjs.api.BundlableNode;
+import org.bladerunnerjs.api.memoization.MemoizedFile;
+import org.bladerunnerjs.api.model.exception.ModelOperationException;
+
+
+public class DirectoryAsset implements DirectoryLinkedAsset
```
dchambers: Is `DirectoryAsset` the replacement for `AssetLocation`?

brjs-core-tests/src/test/java/org/bladerunnerjs/spec/aliasing/AliasModelTest.java
```diff
@@ -1,332 +0,0 @@
...
-import org.bladerunnerjs.api.model.exception.request.ContentFileProcessingException;
-import org.bladerunnerjs.api.spec.engine.SpecTest;
-import org.bladerunnerjs.model.AssetContainer;
-import org.bladerunnerjs.utility.FileUtils;
-import org.junit.Before;
-import org.junit.Test;
-
-import com.ctc.wstx.exc.WstxValidationException;
-
-public class AliasModelTest extends SpecTest {
```
dchambers: Is there a replacement for this test class since the tests themselves still all seem broadly relevant?

andyberry88: No, but you're right they should at least be converted to test the `AliasingUtility` class or similar. Will look at this today.

brjs-core/src/main/java/org/bladerunnerjs/model/DirectoryAsset.java
andyberry88: Yep.

brjs-core-tests/src/test/java/org/bladerunnerjs/plugin/commands/core/HelpCommandTest.java
andyberry88: No, I don't know where it came from. I'll take a look at it.

dchambers:
Hi @andyberry88, I've just checked the latest diff and there still seems to be a `utility_copy` package around? Additionally, this pull-request can't currently be merged.

brjs-core/src/main/java/org/bladerunnerjs/api/plugin/AssetPlugin.java
```diff
@@ -1,16 +1,23 @@
...
+public interface AssetPlugin extends Plugin
+{
+	/**
+	 * Get the priority for this plugin. Lower priorities correspond to earlier execution.
+	 * The default priority for plugins is 10.
+	 * Any plugin that can detirmine with 100% confidence it can handle a given resource should define a lower priority so it is executed first and can better influence the discovery of assets.
+	 * 
+	 * @return the priority for this plugin. Lower priorities correspond to earlier execution.
+	 */
+	int priority();
```
dchambers: I think this is worse than what we had before as it will no longer always be possible for plug-ins to ensure that they will appear in the correct order relative to other plug-ins, depending on the `priority` values those other plug-ins have chosen for themselves.

Instead, it probably makes more sense to have the plug-in order determined globally within `brjs.conf`, for example:

```
pluginOrder: Plugin1, Plugin2, *, PluginN
```

brjs-core/src/main/java/org/bladerunnerjs/api/spec/engine/ModelCommander.java
```diff
@@ -16,7 +16,7 @@ public ModelCommander(SpecTest modelTest) {
 			specTest.logging.stopStoringLogs();
 		}
 		catch(Throwable e) {
-			if (specTest.catchAndVerifyExceptions) 
+			if (specTest.catchAndVerifyExceptions && !(e instanceof NullPointerException))
```
dchambers: @andyberry88, did you mean to check this in?

dchambers:
Okay, I've finally finished reviewing this epic commit &mdash; great work getting it all done!! For the time being though, moving back to `dev` due to outstanding issues:

  1. It's still unmergable.
  1. The plug-in ordering is not right &mdash; I'd be happy to create a separate issue to address this though, as long as it gets done immediately so that we don't release with an interim API.
  1. Small question about the `&& !(e instanceof NullPointerException)` piece of code.

andyberry88:
:+1: Thanks @dchambers. 

> It's still unmergable.

I'll fix this ASAP so we can get this merged in.

> The plug-in ordering is not right

Agreed. I'll look at that. Either by allowing ordering to be set in `brjs.conf` or moving the list of loaded plugins into `brjs.conf` too and making the ordering part of that declaration.

> Small question about the && !(e instanceof NullPointerException) piece of code.

This was intentionally left in. Given NPE aren't exceptions we'd intentionally throw it's useful to have them rethrown so the stack trace is immediately available in tests.

dchambers:
> This was intentionally left in. Given NPE aren't exceptions we'd intentionally throw it's useful to have them rethrown so the stack trace is immediately available in tests.

:+1: Very good point!

andyberry88:
> The plug-in ordering is not right

This is fixed in #1313. Plugins can now be enabled and ordered via `brjs.conf`.

PR #1278, opened 2015-03-12
Added internationalization for jQuery datepicker

jeremyherr:
Instead of using jQuery datepicker's standard way of configuring
   different locales, we use the default locale, setting locale properties
   to i18n keys, so that the user can internationalize the datepicker
   along with the rest of their app using the BRJS standard *.properties
   files.

To set the datepicker defaults, user needs to add to their app:
```
var JQueryDatePickerLocales = require('br/presenter/control/datefield/JQueryDatePickerLocales');
JQueryDatePickerLocales.setLocalePropertiesToI18n();
```
   Putting this code in the control doesn't make sense because it affects
   all datepickers in the app, including ones not handled by
   JQueryDatePickerControl, and also it would get run every time a new
   control is instantiated.

Maybe there's a better place to call this method than in the user's app?


dchambers:
Hi @jeremyherr, thanks for the commit :+1:

I think that `JQueryDatePickerLocales.setLocalePropertiesToI18n()` should be renamed to something like `JQueryDatePickerLocaleUtil.setDefaultJQueryDatePickerLocaleProperties()`, to make it clearer what's happening.

Or, alternatively, it could be converted into a getter, leaving the end-developer to invoke `jQuery.datepicker.setDefaults()` themselves, for example:

```javascript
jQuery.datepicker.setDefaults(JQueryDatePickerLocaleUtil.getDefaultDatePickerLocaleProperties());
```

which makes the intent of the library even clearer.

bit-shifter:
@dchambers - I agree with the suggested improvement to make the intent more explicit.

It will require a further app-level require, but I don't see this as a bad thing as it further enforces intent.

Are you able to make the change directly?

dchambers:
@bit-shifter: yes, I'll look to start that within the hour...

dchambers:
I actually can't make any changes to this until after it's been merged into develop since the code is in a fork, so moved to `ReadyForTest`.

PR #1281, opened 2015-03-16
Close underlying reader when closing XMLSiblingReader instances.

### LINKED ISSUE #446: BRJS prevents me saving/deleting files (locked files/folders)
I'm not 100% sure this is BRJS's fault, but because I haven't experienced the same issue before with the same workflow I'm fairly certain it is it's fault :)

If I have BRJS running (`brjs serve`) and edit and save a file I will sometimes get an error in my editor (Sublime on Win7) saying that it couldn't save the file because of permission error.
If I then hit save one more time (or twice) the error goes away and the file is saved.

@dchambers has a suspicion that BRJS is not closing down file handles properly or in time.

dchambers:
This check-in potentially fixes #446. Since we've never been able to replicate the problem in the first place (only @janhancic can), it's impossible to be sure though.

### DIFF
--- a/plugins/brjs-plugins/src/main/java/org/bladerunnerjs/plugin/bundlers/xml/XMLContentPlugin.java
+++ b/plugins/brjs-plugins/src/main/java/org/bladerunnerjs/plugin/bundlers/xml/XMLContentPlugin.java
@@ -1,5 +1,6 @@
 package org.bladerunnerjs.plugin.bundlers.xml;
 
+import java.io.IOException;
 import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


ioanalianabalas:
Looks good!

@andyberry88, could you double-check this one, please?

andyberry88:
:+1: All good.

thanhc:
Followed the steps in #446 but not able to reproduce the xml file locking issue, just checked that everythings still working

sospirited:
Might be some a regression here in the latest BRJS where I have on disk:
```
<app>/blades/<bladeName>/resources/xml/rendererDefinitions.xml
```

Occasionally, whilst `brjs serve` is running, attempting to switch branches, git complains that the above file is locked, killing brjs and telling git to attempt this again works.

Will try and find a reliable way to reproduce this.

andyberry88:
@sospirited is the `<app>/blades/<bladeName>/resources/xml/rendererDefinitions.xml` file being deleted when you change git branches? We don't/can't support deleting directories since they are opened by the file watcher. If changing branches is deleting the directory it will fail due to locked directories.

PR #1282, opened 2015-03-17
Allow validators to be removed

dchambers:


andyberry88:
Looks good :+1:

janhancic:
This solution feels really awkward. I'd rather we did the original approach that @mklakakilli proposed. Would you be willing to reconsider this?

PR #1286, opened 2015-03-18
LocalisedDateParser fix

charliehw:
We should be using Moment's strict parse flag to avoid loose matches which can lead to unexpected results.

Without this change, the test example "18 Mar 2015" is parsed as "YYYY" to the year 18. Only exact matches are particularly useful.


andyberry88:
@thanhc can you take a quick look at this ASAP so we can merge it in for the 0.15.2 release?

thanhc:
@andyberry88 looks good and merged

andyberry88:
:+1:

PR #1291, opened 2015-03-19
Outofscopeexception for require 1232

### LINKED ISSUE #1232: [feature] BRJS should throw an exception if requiring something you shouldn't
If my blade is requiring something from the `default-aspect`, then that works just fine when running the app. But when I then test my blade, I get an exception saying that it can't resolve the desired require path (which is, I guess, wrong on it's own, it should distinguish when something isn't there and when something isn't allowed to be required).

I would like to get an exception no matter how and where I'm running my code.

ioanalianabalas:
Fixes #1232.


andyberry88:
This is superseded by #1318. I've taken most of the changes from this PR. applied them to develop and then built on top of it for the extra scope enforcement.

PR #1300, opened 2015-03-24
LocalisedDateFormatter fix

charliehw:
Unlike parsers, which return null or undefined if they fail to parse, formatters are usually expected to return a value regardless. Otherwise, invalid inputs can be removed from the view as a result of formatting. If the LocalisedDateFormatter gets an input that doesn't match the input format, it should just return the input value.


thanhc:
@andyberry88 I accidentally merged this, its all working and change is  minimal but still be good for you to do a quick review.

PR #1304, opened 2015-03-25
BRLocaleProvider modifying read only array

charliehw:
On line 82 this read only array is being modified. This throws an exception in Firefox. We can clone the array with slice(0) to prevent this issue.


andyberry88:
This was added to the 0.15.3 release as ce3be2f94dcf091f38a4b22cdb00acd71c3a48d5 since `master` didn't contain the `BRLocaleProvider` file at that point in the codebase.

PR #1305, opened 2015-03-25
If the Mime type is not found, we continue the search and ignore that fi...

### LINKED ISSUE #1295: import-app throwing jmimemagic exception
While trying to import fxtrader I'm getting the following error:
Error:
  Failed to import application from zip '../../../../FxTraderApp-4.0.3-0-g957197c.zip'.; net.sf.jmimemagic.MagicMatchNotFoundException

Caused By:
  net.sf.jmimemagic.MagicMatchNotFoundException

Stack Trace:
org.bladerunnerjs.api.model.exception.command.CommandOperationException: Failed to import application from zip '../../../../FxTraderApp-4.0.3-0-g957197c.zip'.; net.sf.jmimemagic.MagicMatchNotFoundException
        at org.bladerunnerjs.plugin.commands.standard.ImportAppCommand.doCommand(ImportAppCommand.java:83)
        at org.bladerunnerjs.api.plugin.ArgsParsingCommandPlugin.doCommand(ArgsParsingCommandPlugin.java:44)
        at org.bladerunnerjs.plugin.proxy.VirtualProxyCommandPlugin.doCommand(VirtualProxyCommandPlugin.java:41)
        at org.bladerunnerjs.utility.CommandRunner.run(CommandRunner.java:25)
        at org.bladerunnerjs.api.BRJS.runCommand(BRJS.java:406)
        at org.bladerunnerjs.utility.UserCommandRunner.doRunCommand(UserCommandRunner.java:34)
        at org.bladerunnerjs.utility.UserCommandRunner.run(UserCommandRunner.java:26)
        at org.bladerunnerjs.api.BRJS.runUserCommand(BRJS.java:411)
        at org.bladerunnerjs.runner.CommandRunner.run(CommandRunner.java:112)
        at org.bladerunnerjs.runner.CommandRunner.main(CommandRunner.java:64)

ioanalianabalas:
...le.

FIxes #1295.

### DIFF
--- a/brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
+++ b/brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
@@ -205,21 +205,29 @@ public class NodeImporter {
 	{
 		for (File f : files) {
 			if (f.length() != 0) {
-				if (checkFileMimeType(f).startsWith("text")) {
-					findAndReplaceInTextFile(brjs, f, sourceRequirePrefix, targetRequirePrefix);
+				try {
+					if (checkFileMimeType(f).startsWith("text")) {
+						findAndReplaceInTextFile(brjs, f, sourceRequirePrefix, targetRequirePrefix);
+					}
+				}
+				catch (MagicMatchNotFoundException e) {
+					continue;
 				}
 			}
 		}
 	}
 
-	private static String checkFileMimeType(File file) throws IOException {
+	private static String checkFileMimeType(File file) throws IOException, MagicMatchNotFoundException {
 		byte[] data = Files.readAllBytes(Paths.get(file.getAbsolutePath()));
 		MagicMatch match = null;
 		try {
 			match = Magic.getMagicMatch(data);
-		} catch (MagicParseException | MagicMatchNotFoundException
-				| MagicException e) {
+		} 
+		catch (MagicParseException | MagicException e) {
 			throw new IOException(e);
+		}
+		catch (MagicMatchNotFoundException e) {
+			throw e;
 		}
 		return match.getMimeType();
 	}


brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
```diff
@@ -205,22 +205,30 @@ private static void findAndReplaceInTextFiles(BRJS brjs, Collection<File> files,
...
 		for (File f : files) {
 			if (f.length() != 0) {
-				if (checkFileMimeType(f).startsWith("text")) {
-					findAndReplaceInTextFile(brjs, f, sourceRequirePrefix, targetRequirePrefix);
+				try {
+					if (checkFileMimeType(f).startsWith("text")) {
+						findAndReplaceInTextFile(brjs, f, sourceRequirePrefix, targetRequirePrefix);
+					}
+				}
+				catch (MagicMatchNotFoundException e) {
```
andyberry88: Is this not going to prevent property files etc. from being filtered? Maybe we should filter if theres a `MagicMatchNotFoundException`.

brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
```diff
@@ -205,22 +205,30 @@ private static void findAndReplaceInTextFiles(BRJS brjs, Collection<File> files,
...
+					continue;
 				}
 			}
 		}
 	}
 
-	private static String checkFileMimeType(File file) throws IOException {
+	private static String checkFileMimeType(File file) throws IOException, MagicMatchNotFoundException {
 		byte[] data = Files.readAllBytes(Paths.get(file.getAbsolutePath()));
```
andyberry88: This may become very inefficient. It might be worth not checking based on mime type and having a list if known extensions we filter. While it's a list we have to maintain we won't need to read every file in the app to figure out whether we need to read it again and replace the content.

brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
dchambers: According to [this slashdot article](http://stackoverflow.com/questions/6248130/what-is-the-mime-type-for-properties-files), the content type is likely to be either `text/plain` or `text/x-java-properties`, either of which would work fine.

brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
dchambers: @andyberry88, I think your idea is sound, but I don't think it's particularly high priority since [jmimemagic](https://github.com/arimus/jmimemagic) is actually quite efficient in that it only scans the beginning of the file, and given that importing apps is only performed occasionally and is not a time critical operation.

thanhc:
I've testing importing an app with .properties files and it is being re-namespaced correctly

andyberry88:
Since time is tight for the next release we can merge this in and discuss the points above after the release.

brjs-core/src/main/java/org/bladerunnerjs/model/NodeImporter.java
andyberry88: > but I don't think it's particularly high priority since jmimemagic is actually quite efficient in that it only scans the beginning of the file, and given that importing apps is only performed occasionally and is not a time critical operation.

Agree on the low priority part. We are reading the entire file though before passing it to the mime type library (`byte[] data = Files.readAllBytes(Paths.get(file.getAbsolutePath()));`) so it is going to be much slower for large apps.

andyberry88: The use of MimeMagic was the cause of #1315 since files with `£` etc. aren't seen as text files. The root cause of the need for MimeMagic was a String == String comparison which meant we always wrote to the file even if we hadn't matched any namespaces to change. This has now been fixed so we can remove the need for MimeMagic and do an initial check for known file types, to prevent reading extra image files we know won't change, and then only write to a file where we need to change the contents.

PR #1309, opened 2015-03-27
Handle arrays in MapUtility#deepClone

### LINKED ISSUE #1301: MapUtility#deepClone doesn't clone arrays
The `MapUtility#deepClone` function doesn't properly clone a map that has an array in it. The array gets converted into an object.

The funny thing is that the docs for `deepClone` actually say: 

> (note that non-map objects are not handled correctly)

Which probably means we can't just fix this without potentially breaking backwards compatibility as some people might be relying on the current behaviour.

@james-shaw-turner what do you think? Is it low risk enough that we can just fix it or not?

janhancic:
Before it would convert Arrays into objects. Now it creates a new array and then recursively clones each element in that array.

This fixes #1301

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-util/src/br/util/MapUtility.js
+++ b/brjs-sdk/sdk/libs/javascript/br-util/src/br/util/MapUtility.js
@@ -334,8 +334,7 @@ MapUtility.clone = function(srcMap) {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


dchambers:
Thanks @janhancic :+1:

ccpetercc:
Closing pull request: see updated - #1322

PR #1311, opened 2015-03-30
Adding brjs-apps to dirs watched by file observers and observer API changes

ioanalianabalas:


andyberry88:
We should probably make sure the new polling thread also watches `brjs-apps` in the pull request too.

andyberry88:
I've merged several changes in to this PR:
- add `brjs-apps` to the polling file observer too
- change the file observer API so it has `start()` and `stop()` methods
- remove the extra debugging code (see my comment above)

andyberry88:
Following suggestions from the sprint demo and my comment on https://github.com/BladeRunnerJS/brjs/issues/1154#issuecomment-88483893 I've added more commits to this PR to not require `brjs-apps` to be present if an app is found as we traverse up the directory tree.

ioanalianabalas:
Dev-reviewed: looks good!

thanhc:
Not quite working, it seems like when running brjs within an app its treating the app folder as brjs-apps, so it needs to look one directory up.
Also should it only display actual apps in the dashboard? right now when i run brjs serve from in the app the dashboard shows me each folder as an app, but surely it should only display actual apps?

andyberry88:
> Not quite working, it seems like when running brjs within an app its treating the app folder as brjs-apps, so it needs to look one directory up.
Also should it only display actual apps in the dashboard? right now when i run brjs serve from in the app the dashboard shows me each folder as an app, but surely it should only display actual apps?

The wrong apps folder was being detected due to the working dir file path ending with `.` which meant parent file was the working dir. Fixed in c6277c694509735cb3aa4c888c3262991421e3bf.

PR #1313, opened 2015-03-31
Allow plugins to be ordered via brjs.conf

andyberry88:
PR for #1085..
Allowing the ordering of plugins and the plugins that are enabled to be ordered via `brjs.conf`.


dchambers:
@andyberry88, again, would I be right in assuming that only the last four commits (this time 765ae43, fb002eb, 5324614 & c3383c7) need reviewing?

andyberry88:
Yep, only the last few commits need reviewing since this branch was based on the API changes branch.

dchambers:
Apart from the fact that this pull-request has merge conflicts, this is ready for testing.

andyberry88:
Moving this to `ReadyForTest`. I'll find the conflicts once it's been QA'd otherwise we'll just end up with more conflicts as things are merged into `develop`.

dchambers:
@thanhc just mentioned how this change has led to us having a lot more configuration within `brjs.conf`, which made me remember that most plug-ins don't need ordering, so most of this new config is superflous.

However, a recent change (1a235bd) has caused all plug-ins to become ordered even though they don't need to be. Can we instead go back to Having an `OrderedPlugin` sub-interface so that we can again only have ordering for plug-ins that need it?

ccpetercc:
I've tested this and the ordering defined in brjs.conf is applied successfully

I've moved the issue back to dev whilst Dom's question is discussed.

ccpetercc:
Currently, if a non-existent plugin is added to the brjs.conf file then no error is thrown and dashboard and applications are served successfully. Is this expected behaviour?

andyberry88:
> Can we instead go back to Having an OrderedPlugin sub-interface so that we can again only have ordering for plug-ins that need it?

Allowing plugins to be defined doesn't just allow ordering, but also enabling/disabling plugins through the same mechanism. I'd suggest that we fix the issue that means we can't have defaults in YAML files without them being written to `brjs.conf`. e.g. any public field in the class that represents the YAML file is written out to the conf file, no fields can be hidden.

andyberry88:
> Currently, if a non-existent plugin is added to the brjs.conf file then no error is thrown and dashboard and applications are served successfully. Is this expected behaviour?

Yes. There should be a warning logged for the BRJS console though. This is so that if a plugin is listed in `brjs.conf` and isn't available it doesn't prevent everything from working as normal, albeit without that plugin.

dchambers:
> Allowing plugins to be defined doesn't just allow ordering, but also enabling/disabling plugins through the same mechanism. I'd suggest that we fix the issue that means we can't have defaults in YAML files without them being written to brjs.conf. e.g. any public field in the class that represents the YAML file is written out to the conf file, no fields can be hidden.

Apps generally only make use of the plug-ins they want to use anyway &mdash; for example, an app must use tags to use any of the available tag-handler plug-ins, or make requests to the server to use any of the content plug-ins, etc. Model observer plug-ins are the only plug-in type where that doesn't seem to be the case, and these are mostly harmless anyway. Perhaps you can provide a good use case of plug-in we have that an app might want to disable completely?

Additionally, `brjs.conf` is not a good place for app-specific plug-in settings anyway, since any settings here apply equally to all applications.

PR #1317, opened 2015-04-01
Renamespacing when the namespace is preceded by a slash or the file contains £

### LINKED ISSUE #1315: Renamespaced content overwrites . seperators with / & doesnt renamespace files with a £
I'm having two issues when importing an app using 0.15.3. In one case, a test js file in the app is not having namespaced references replaced to the app's namespace so the tests fail. Other test files that exist in the same folder are having their references namespaced. I can't see how the first file is different to the others.

The second issue occurs when namespacing some xml that exists in a test file as a string:
    `"<caplinx.fxexecution.tilelayout.tile-set></caplinx.fxexecution.tilelayout.tile-set>"`
becomes
    `"<novo.fxexecution.tilelayout.tile-set></novo/fxexecution.tilelayout.tile-set>"`

The double / in the output causes a parse error and the tests fail.

I have sent the app I'm struggling with to @ioanalianabalas

ioanalianabalas:
Fixes #1315.

When BRJS finds that the matched namespace / require path contains the character '/', it replaces it with the corresponding require path if it otherwise does not contain '.'. If '.' are detected after the first character, the new namespaced it used.


ccpetercc:
I've tested the renamespace-with-separators-1315 branch running gradlew brjs-sdk:distZip and using that as a dependency in CT (branch upgrdage-brjs-1.0). I then imported Charlie's fxtrader App which causes the bugs (using Import Motif from ZIP) setting the namespace to fxtrader2.

The issue in /ct-core/apps/fxtrader2/fxexecution-bladeset/blades/blocktrade/tests/test-unit/js-test-driver/tests/upload/presentation/ExcelImportControlTest.js regarding the £ character does not appear to have been fixed - 

After importing

line 7 is:
fields: caplinx.fxexecution.blocktrade.manager.presentation.Manager.prototype._generateExcelFields()
when it should be:
fields: fxtrader2.fxexecution.blocktrade.manager.presentation.Manager.prototype._generateExcelFields()

Having set the namespace to fxtrader2


The issue in /ct-core/apps/fxtrader2/fxexecution-bladeset/blades/tilelayout/tests/test-unit/js-test-driver/tests/fxtrader2/fxexecution/tilelayout/TileSetGroupTest.js

does appear to have been fixed. The XML is now correct.

I wasn't able to test this extensively. Errors are thrown by Charlies fxtrader app and it is not possible to run successfully. This is most likely because I imported the App into the CT branch, upgrade-brjs-1.0, which is required to test the BRJS SDK dependency.

andyberry88:
I've added an extra commit to this to fix the re-namspacing files with `£` chars in them. It was caused by the MimeMagic utility we were using - see my comment for details on this change https://github.com/BladeRunnerJS/brjs/pull/1305#discussion_r27649593.

Issue #565, opened 2014-04-23
As a developer, when my require() fails, I want to know the locations BRJS attempted to find my dependency


sospirited:
There are multiple locations/scopes from where dependencies can be required.
E.g.
- sdk/libs/javascript/br-libs
- sdk/libs/javascript/thirdparty
- `<appName>`/default-aspect/src/
- `<appName>`/`<myBladeset>`/`<myBlade>`/src
- `<appName>`/libs
etc

This will help developers better understand why certain items aren't being bundled when invalid requirePaths are made (e.g. bladesets trying to require a blade class)

If we're failing fast anyway, the above information would make for such a richer debugging experience and reduce dev frustration!

dchambers:
This issue will probably be best fixed by using @janhancic's proposal in #1232.

dchambers:
Fixed by #1291.

ccpetercc:
For #1291:
Also need to add a message saying which file the guilty require appears in

andyberry88:
#1318 is now the PR for this. I've taken most of the changes from #1291. applied them on top of the API changes in #1277 and then built on top of it for the extra scope enforcement in #1232.

andyberry88:
While there are some issues relating to #1232 this particular issue has been fixed in #1318 and IMO can be closed once that is merged.

PR #1318, opened 2015-04-02
OutOfScopeExceptions that list the scope and stricter enforcement so Blades cant depend on Blades in the Aspect scope

### LINKED ISSUE #1232: [feature] BRJS should throw an exception if requiring something you shouldn't

andyberry88:
Fixes #1232 and #565.

- The out of scope exception now lists the valid scopes for dependencies
- Scope enforcement is stricter - the location of the dependant file is checked to ensure it's within the scope of the source file in addition to being in the current bundlable scope

I've had to add a `Asset.isLogicalAsset()` so we can determine whether the enforcement above can be applied as things like Alias assets can depend on assets outside of the scope and the Alias Registry depends on AliasData which is outside of it's scope. @dchambers or @james-shaw-turner can you review this? 

This is based on the API changes branch (#1277) and should be merged in after that PR.


dchambers:
@andyberry88, am I right in assuming that only the last four commits (#d8c9c60, #482f653, #0cb20d7 & #3646226) needed dev-reviewing?

That being the case, this has now been reviewed...

andyberry88:
Yep, only the last few commits need reviewing since this branch was based on the API changes branch.

PR #1321, opened 2015-04-07
ensuring jetty-env is renamespaced propertly

andyberry88:
- fixing the find/replace logic for the app name in jetty-env
- making sure we use the BRJS conf file encoding when reading in files to re-namepsace them

This is needed for the 0.15.4 release. @dchambers can you dev review ASAP so @thanhc can test and merge?


thanhc:
all good!

PR #1323, opened 2015-04-07
Fix legacy clear patch

### LINKED ISSUE #1213: legacyClear() patch doesn't work in practice
@janhancic has seen an issue that happens when:

  1. CommonJs modules that rely on service(s) are being tested.
  1. The `legacyClear()` patch is used to cause the `ServiceRegistry` to be automatically cleared down after each test is run.

which effectively means that the patch doesn't work.

We could work around this by:

  1. Changing our js-test-driver code to use sub-realms or invoke a new `realm.clear()` method at the same time it invokes `serviceRegistry.clear()` (for developers that still have the patch).
  1. Giving up on the patch idea, and converting all tests that rely on the patch to manually invoke `serviceRegistry.clear()`.

The problem with the first option is that tests will become slower as a result, since clearing the service registry is much cheaper than clearing every single module. The problem with the second idea is that it will require existing clients to change lots of tests.

Some thoughts:

  * Perhaps, making it reliable but slow will be a catalyst for people removing the patch?
  * Perhaps it won't be that slow after all, since I'm just making an unfounded assumption.
  * I may have got some of the details as to when this happens wrong (@janhancic wondered if it depended on having NamespacedJs tests for example), so it may be more of an edge case than I've made out (this is after all the first anybody has seen the problem).

@andyberry88 / @james-shaw-turner / @bitshifter: thoughts?

dchambers:
This fixes #1213.


dchambers:
This is superseded by #1324.

PR #1325, opened 2015-04-08
Add support for OptionsNodeList in popup windows

dogoku:
Fixed a bug where OptionsNodeList would not generate a correct array of options when running in a popup windows.

Arrays coming from main window will fail "vOptions instanceof Array" checks.
Instead use toString to check for "[object Array]"


ioanalianabalas:
Dev reviewed - looks good!

Issue #1290, opened 2015-03-19
Allow a meaningful version tag to be used


dchambers:
At present we version bundle URLs so that browsers are able to indefinitely cache an app's content's. The version identifier we use is either `dev` (in development) or an epoch time like `1426768892800` (in production). Although epoch times are efficient in-memory representations of dates, they are no more efficient to store as strings than ISO style dates, which are easier for administrators to comprehend.

To improve things we'd like to:

  1. Use an easier to comprehend date format like `20150319124132` rather than `1426768892800` (only 1 character more).
  2. Add a `--version-prefix` flag to the `build-app` command that allows a meaningful version prefix (e.g. `3.5.2`) to be provided, causing a composite version of `3.5.2.20150319124132` to be used.

This proposal solves a number of problems:

  * It allows BRJS applications to know what version they are.
  * It allows system administrators looking at web access logs to know both the version number and the build time-stamp of the app being requested.
  * It does this while continuing to prevent people from accidentally generating apps having different contents, but the same version identifier.

dchambers:
@janhancic, is this proposal acceptable for your needs?

janhancic:
Looks good. What would be a reliable way for developers to split that version provided by BJRS into the version-prefix and the timestamp? Would there be a delimiter?

janhancic:
Tagging @adamshone and @sospirited

bit-shifter:
On the face of it, the proposal looks like it might resolve a client support issue too.

stevesouth:
Please consider the work that has already been done with the appcache plugin. We have an appcache.conf file that has two properties, version and devVersion e.g.

appcache.conf
version: 1.1.2-464

and if you want to enable appcache in dev (which you would only do if you were developing a version managment UI), then you can do:

devVersion: 1.0.0

Happy to have a session to talk through how we manage versions on the mobile project. It is probably more than we need for desktop apps, but at we should try and converge on a versioning solution. Perhaps we simply need to move 'version' and 'devVersion' into app.conf, but there may be other options.

dchambers:
Thanks @stevesouth, it sounds like we need to move `version` and `devVersion` into `app.conf`, and start automatically making use of those properties if they are available, for example:

  * The `build-app` command would use the `version` config if available rather than having a `--version-prefix` flag.
  * The `serve` command would use the `devVersion` config if available rather than using a version of `dev`, allowing caching to be tested in dev.

janhancic:
[...]

PR #1327, opened 2015-04-09
Allow the version to be configured

andyberry88:
Allows the app version to be configured via a `--version` flag in the `serve` and `build-app` commands. The full App version is the version supplied via the command followed by a timestamp to ensure the version is unique which avoids caching problems.

PR for https://github.com/BladeRunnerJS/brjs/issues/1290.


andyberry88:
Closing this and re-creating it as it's against `master` rather than `develop`.

PR #1328, opened 2015-04-09
Allow the version to be configured

andyberry88:
Allows the app version to be configured via a `--version` flag in the `serve` and `build-app` commands. The full App version is the version supplied via the command followed by a timestamp to ensure the version is unique which avoids caching problems.

PR for https://github.com/BladeRunnerJS/brjs/issues/1290.


andyberry88:
@dchambers can you dev review this please? If it's an acceptable solution for #1290 we can get it in to the 1.0-RC1 release.

brjs-core/src/main/java/org/bladerunnerjs/model/AppVersionGenerator.java
```diff
@@ -3,7 +3,9 @@
 
 public interface AppVersionGenerator
```
dchambers: It's not clear to me why we need `dev` and `prod` versions? Isn't it the case that the version string just tends to have a different form when run in `dev` and `prod`?

andyberry88:
@dchambers Agree with everything above. I've fixed all these in d1441d20e5bee146c480322fdbc9bb6468956483.

dchambers:
Yep, looks good @andyberry88.

thanhc:
just a couple of issue, the timestamp should be 24hrs format and ideally add timestamp to dev version when versioned

PR #1332, opened 2015-04-13
adding an aliasDataPlugin that returns alias data for the BundlableNode

andyberry88:
- adds an AliasData require plugin that returns the alias data for the BundlableNode instead of trying to find it from scope asset containers
- prevents ambiguous require path exceptions and from returning the asset data for a different bundlable node that the one we are using
- should also fix the current failures in `develop`

### DIFF
--- a/plugins/brjs-plugins/src/main/java/org/bladerunnerjs/plugin/require/AliasDataRequirePlugin.java
+++ b/plugins/brjs-plugins/src/main/java/org/bladerunnerjs/plugin/require/AliasDataRequirePlugin.java
@@ -1,0 +1,40 @@
+package org.bladerunnerjs.plugin.require;
+
+import org.bladerunnerjs.api.Asset;
+import org.bladerunnerjs.api.BRJS;
+import org.bladerunnerjs.api.BundlableNode;
+import org.bladerunnerjs.api.model.exception.RequirePathException;
+import org.bladerunnerjs.api.plugin.RequirePlugin;
+import org.bladerunnerjs.api.plugin.base.AbstractRequirePlugin;
+import org.bladerunnerjs.plugin.plugins.require.AliasDataSourceModule;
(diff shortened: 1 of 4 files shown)


andyberry88:
@ioanalianabalas or @dchambers can you dev review this so it can go into `ReadyForTest` so @thanhc can QA it ready for the 1.0-RC1 release?

plugins/brjs-plugins/src/main/java/org/bladerunnerjs/plugin/require/AliasDataRequirePlugin.java
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	@Override
+	public String getPluginName()
+	{
+		return "alias";
+	}
+
+	@Override
+	public Asset getAsset(BundlableNode bundlableNode, String requirePathSuffix) throws RequirePathException
```
dchambers: This method seems to be the only non-test method of any consequence, yet it doesn't seem to seem to do very much, so that I'm struggling to reconcile this method with the claimed commit message &mdash; `adding an aliasDataPlugin that returns alias data for the BundlableNode`.

Can you provide some more detail?

andyberry88: @dchambers We were previously getting `AmbiquousRequirePath` exceptions because `alias!$data`) is registered for each `BundlableNode` and the scope for an aspect bundle includes 2 bundlable nodes (the Aspect and the test pack). The alias data plugin ensures that when the alias data is required only the alias data for the current bundlable node is returned.

thanhc:
@andyberry88 there's still one failing test in the exampleAppsTest its in the it-app peter created regarding the 3rd party libs.

PR #1336, opened 2015-04-13
Fix IE LocaleSwitcher bug

dchambers:
@thanhc, I think this may fix the bug you've been seeing on Jenkins.


dchambers:
@thanhc, it turns out that this isn't the only IE8 bug at the moment. I've been investigating IE8-IE10 failures (in presenter ATs) that I'd assumed that I'd caused, but it turns out these were actually introduced somewhere between `0.15.4` and `develop`.

Still investigating...

andyberry88:
@dchambers I haven't had chance to have a proper look at the current failures in IE9 & IE10 but it may be related to the jQuery/Knockout/test utils issue I saw a few weeks ago where Knockout is using jQuery because it's there and globalised but our test utils don't use jQuery for events.

dchambers:
@andyberry88, yes you were right. This has now been fixed in #1337.

thanhc:
looks good, test are now passing in IE8

PR #1340, opened 2015-04-15
Review plugin apis improvements

andyberry88:
fixes for the last few bugs following the API improvements, out of scope exceptions & brjs from anywhere changes
- improving the out of scope messages
- only detecting dirs with app.conf, index.html, WEB-INF etc. as apps (##1331)
- the stats plugin now tracks the correct bundleset generation time
- aliases and services aren't included in NamespacedJS `requireAll` calls
- performance improvements - the BundleSet caches source modules of each type as they are requested


ioanalianabalas:
Dev-reviewed. :+1:

PR #1349, opened 2015-04-17
Performance improvements after API changes

andyberry88:
Minimising the amount of extra Aliasing work we have to do by using memoized values.
@dchambers can you dev review this ASAP? This brings performance back in line with v0.15.x releases and is needed for the 1.0-RC1 release.


ioanalianabalas:
@andyberry88, I have dev reviewed this and commented on a couple of `try` blocks where we are explicitly closing the stream.

andyberry88:
@ioanalianabalas Good spot, I didn't realise `ResponseContent` extends `AutoClosable`, so you're right we don't need to explicitly close the response. This is fixed in a23ad4c.

dchambers:
@andyberry88, FYI, if you can put the assignment inside the `try` brackets then you know it's auto-closable, since it will error otherwise.

PR #1351, opened 2015-04-20
xmlns changes

### LINKED ISSUE #1345: change xmlns fields to bladerunnerjs.org URIs
We currently use schema.caplin.com to host alias schemas (see below.) These should be moved into a subfolder of bladerunnerjs.org.

```
<aliases xmlns="http://schema.caplin.com/CaplinTrader/aliases">
	<!-- you can add any overrides you need here -->
</aliases>
```

andyberry88:
changing any instances of schema.caplin.com to schema.bladerunnerjs.org - fixes #1345 

This depends on #1349 and should be merged in after that.


andyberry88:
I've made changes so the old `xmlns` attributes will continue to work with a warning logged.

Issue #1116, opened 2014-12-02
Make the HTML resource service work for all known use cases


dchambers:
We've changed `BRHtmlResourceService` a number of times to fix bugs, where each time it's caused some other problems. We need to understand the complete set of use-cases and issues that need to be addressed, and tackle this holistically. This issue should be used to collate all of this information in one place before we contemplate making any changes.

The _resource-service_ should:

  * Allow the developer to place templates within a `<template/>` tag, but continue to support arbitrary tags with an `id` attribute.
  * Allow any tag to be used as the top-level node (e.g. `<td>`) and still work on IE8.
  * Place the templates within a `<template/>` tag on browsers where this is supported, so that web components can be used.
  * Provide a clone of the template when `getHtmlTemplate()` is invoked (rather than relying on Presenter or Knockout to do this).

Additionally, the _resource-service_ should continue to support the following use cases:

  * We should be able to support templates that have been defined within a `<script/>` tag.
  * We should be able to support templates that contain non-standard (e.g. HTML5) tags.
  * Templates should't be added to the document, or should be added in a way that won't interfere with any web application.

**UPDATE -- FINAL PLAN:**

Mandatory Changes:

  1. The BRJS server will be modified to allow templates to be provided within `<template/>` tags, and to wrap templates that aren't provided within `<template/>` tags, so that all templates provided to the client are wrapped within `<template/>` tags.
  1. In conjuction within the above change, the `HTMLResourceService` will be modified to return a fresh _document-fragment_ of the template contents, rather than providing a single template element (and re-using the same element each time at that).

Optional Post 1.0 Changes:

  1. An optional `<@brjs-templates@>` tag could be made available that can be used to load the HTML templates within the page itself, rather than requiring a separate XHR request.
  1. The `HTMLResourceService` could then be modified to detect if the templates have been pre-loaded, and use these directly instead of loading via XHR.
  1. IE8 support for nested `<td>` elements, etc, could be provided by using a JavaScript HTML parser to convert the `innerHTML` returned for each `<template/>` tag into a `DocumentFragment` instance.

This plan meets all of the originally stated goals, plus one more:

  1. It allows templates that contain more than one element to be created.

Though it does have one minor downside:

  1. `<img>` tags within templates will load the images in the background in IE, though we almost never use `<img>` tags since they are inconvenient to use in BRJS.

dchambers:
I was hoping to use a template polyfill, but learned that:

[...]

PR #1352, opened 2015-04-20
Allow templates to be embedded in index page

dchambers:
This pull-request is dependent on #1338 (first relevant commit is e7286b6). It fixes the first two optional parts of #1116.


dchambers:
@andyberry88, I moved to `ReadyForTest` on the assumption that you have finished dev reviewing this issue?

andyberry88:
@dchambers Yep, just that one suggestion.

thanhc:
there are some test failing in CT with the HTML resource bundling change, ct-workbench and ct-presenter
one of the test failing is:
 Connection Status Editor Workbench Tests. Checks all buttons are available and 'OK' is the initial state
with the following error:
The 'caplin.workbench.connection-editor' template contained more than one root node -- use getTemplateFragment() instead.

thanhc:
In IE we're actually getting more test failures with the same error message regarding template containing more than one node., plus other failures.

In IE8 we're gettign type error's: object doesn't support this property or method

thanhc:
In IE8 when adding the santizeHTML to the innerHTML some template tags aren't being added causing failures in IE8

thanhc:
Found an issue with using <@html.bundle@/> tag when you have a template with multiple templates then we dont seem to fix the auto wrapped templates as a result we get the "template contained more than one root node" error.  I have made changes to the it-app so that itapp.itbladeset.itblade.view has two templates in that file, then added the <@html.bundle@/> tag to the index.html (commented out), the app loads correctly when this is commented out but fails when you uncomment it.  I tried to write the following spec test but it didn't seem to catch the issue:

@Test
	public void multiTemplatesAreLoadedCorrectlyWhenEmbededInIndexPage() throws Exception {
		given(aspect).containsFileWithContents("index.html", "'<html><head><@html.bundle@/></head> hello world  </html>'")
			.and(aspect).containsResourceFileWithContents("html/view.html", "<template id='appns.view'>TESTCONTENT</template> <template id=appns.view-2>BOB</template>");
		when(aspect).requestReceivedInDev("html/bundle.html", response);
		then(response).containsOrderedTextFragments("TESTCONTENT", "BOB");
	}

andyberry88:
@thanhc The issue with the IT-app is fixed in fbd9dd2.

thanhc:
@andyberry88 looks better, the templates are now being separated out correctly., however if in the template I use i18n tokens e.g. <label>@{ct.grid.refine.refinersummary.filters}</label> then these aren't being replaced

andyberry88:
@thanhc The i18n token issue is fixed in the latest commits. We now replace i18n token in embedded templates on the server. If it's loaded via XHR it will continue to be replaced in the client.

PR #1359, opened 2015-04-24
Stop using jsunit extensions

dchambers:
This pull-request does not change any functionality, but reduces our dependence on the `jstestdriverextensions` and `jsunitextensions` libraries where this wasn't actually needed.


dchambers:
If you're dev reviewing you may want to use `?w=1` as my Atom settings at the time were set to 'fix' whitespace.

PR #1361, opened 2015-04-27
accepting 'y' as the default stats plugin answer and adding more tests

andyberry88:
- 'y' is now the default stats plugin answer if no input is provided and adding some more tests around that area of the code


thanhc:
@andyberry88 when running a clean assemble, the brjs.conf is being created with allowAnonymousStats: false.

andyberry88:
@thanhc our build uses the 'no-stats' flag to disable the stats. If you build it using distZip and run a command using the built zip you should see the question asked when BRJS is first run.

PR #1363, opened 2015-04-27
Clock refactoring

ioanalianabalas:
Clock refactoring for stopping breaking Jasmine.

Created the separate Clock class in order to separate it from jsunitextension.

We also added methods for installing and uninstalling the Clock prior to and after use, as well as corresponding tests.


brjs-sdk/sdk/libs/javascript/jsunitextensions/Clock.js
```diff
@@ -0,0 +1,114 @@
...
+				runAtMillis: Clock.nowMillis + millis,
+				funcToCall: funcToCall,
+				recurring: recurring,
+				timeoutKey: timeoutKey,
+				millis: millis
+		};
+};
+
+ClockClass.prototype._verifyInstalled = function() {
+	if(!this.installed) throw new Error("Clock can no longer be used until you've installed it. You should install it in your set-up method, and uninstall it in your tear-down method.");
```
andyberry88: I'm not sure saying "can no longer be used" is quite right. I'd suggest changing this to "Clock cannot be used until you've installed it".

ioanalianabalas:
@andyberry88, thank you for your feedback. I have now updated the error message,

andyberry88:
:+1:

PR #1368, opened 2015-04-30
minor changes to embedded html templates

andyberry88:
- renaming the tag handler
- no longer wrapping the templates in a div incase it's used in the `<head>`
- adding a style tag so templates are inivisible if they are loaded before the CSS


dchambers:
Dev reviewed :+1:

PR #1369, opened 2015-05-02
Cleanup templates

janhancic:
Fix code style in templates so that they conform to the Caplin JS style guide.

Also use the correct way of requiring service in CJS.


ioanalianabalas:
Dev reviewed :+1:

PR #1371, opened 2015-05-06
New bundle path in tests

### LINKED ISSUE #1364: use the new form of bundle paths in tests
we're still using the legacy `<type>.bundle` style bundle paths in tests which sets a bad example. we should change all tests (including example apps) to use the new `bundle.<type>` format (e.g. `bundle.html`, `bundle.xml` etc).

andyberry88:
Fixes #1364.

- replacing the old legacy `xyz.bundle` with `bundle.xyz` format.
- **this is a breaking change** as the JSTD XML and HTML resource services both automatically request the bundles and the URL has now changed. I tried to make it backwards compatible by falling back to the legacy bundle URL but the way JSTD behaves when running multiple test suites made the behaviour nondeterministic, it was also much slower. Because of the backwards breaking change for XML and HTML I've also removed the legacy behaviour for all bundles and removed the dead code in the bundle file creator.

This change will mean instances of the following in all `JSTestDriver.conf` files must change:

```
bundles/xml.bundle -> bundles/bundle.xml OR bundles/xml/bundle.xml
bundles/html.bundle -> bundles/bundle.html OR bundles/html/bundle.html
bundles/js.bundle -> bundles/bundle.js OR bundles/js/dev/<minification-level>/bundle.js
bundles/css.bundle -> bundles/bundle.css OR bundles/css/<theme>/bundle.css
bundles/i18n.bundle -> bundles/i18n/en.js OR bundles/i18n/<locale>.js
```

I've included both the logical path that's replaced each legacy path. These translate to real content paths and the content paths can be used instead, so the format for bundle paths is `bundles/<content-path>. This isn't added in this PR, it's something that's been supported for a while.


dchambers:
Added the `4 - ReadyForTest` label as we have another breaking change that we definitely do want to release (also involving mandatory changes to `jsTestDriver.conf`), and so this becomes as a good a time as we're ever going to have for introducing this breaking change.

brjs-sdk/apps/it-app/login-aspect/test-unit/jsTestDriver.conf
```diff
@@ -3,7 +3,7 @@ basepath: .
 
 load:
   - bundles/i18n/en_i18n.bundle
```
thecapdan: this line is no longer supported and will throw an error

thecapdan:
This needs an extensive release note since it's a breaking change. Probably should include a copy of the table above (which was incorrect) should look like this:

```
bundles/xml/xml.bundle -> bundles/bundle.xml OR bundles/xml/bundle.xml
bundles/html/html.bundle -> bundles/bundle.html OR bundles/html/bundle.html
bundles/js/js.bundle -> bundles/bundle.js OR bundles/js/dev/<minification-level>/bundle.js
bundles/css/css.bundle -> bundles/bundle.css OR bundles/css/<theme>/bundle.css
bundles/i18n/i18n.bundle -> bundles/i18n/<locale>.js
bundles/i18n/<locale>_i18n.bundle -> bundles/i18n/<locale>.js
```

andyberry88:
See https://github.com/BladeRunnerJS/brjs/issues/1364#issuecomment-145802879.

Issue #1374, opened 2015-05-08
Strange bindings within bindingHandler using C3 library


oniseijin:
Similar to another issue about width/height within bindings (perhaps purely KnockoutJS)

Referencing the bottom widget at: http://signaller-rat-84376.bitballoon.com/

Problem: The initial layout has 0 width -> meaning the when binding, it doesn't know the right size. Clicking on the text it displays. However, when rendering purely to a div element (the top of page) it works fine. Unclear where in the toolchain this is getting broken. 

Side problem: C3 doens't resize well within bootstrap, but that is another issue entirely. 

Using the C3 charts library: http://c3js.org/

Code:
```
ko.bindingHandlers.lineChart = {
	init : function(element, valueAccessor){
		c3.generate({
			bindto: element,
			//data: ko.unwrap(valueAccessor()) //not accepted

			data: {
				columns: [
					['data1', 30, 200, 100, 400, 150, 250],
					['data2', 50, 20, 10, 40, 15, 25]
				]
			}

		});
	},

	update: function(element, valueAccessor){

	}
}
```

Template:
```
<div id="kendoui.c3charts.view-template">

	<div id = "c3chart"></div>
	<div data-bind="lineChart: dataset"></div>

</div>
```

Workbench screenshot:
![workbench2](https://cloud.githubusercontent.com/assets/80143/7547509/8f3e6526-f629-11e4-9e07-0806f43fe0ff.png)

oniseijin:
@dchambers  on your advice, this does work, although, would take a bit more work on my end to make the c3 chart reactive to data: 
```
'use strict';

var br = require('br/Core');
var ko = require( 'ko' );
var i18n = require( 'br/I18n' );
var Component = require('br/component/Component');
var c3chart = require("c3");

function C3Component() {
	this.m_oFrame; 
	this._rootElement = document.createElement('div');
	this.getElement().style.width = "300px";
	this.getElement().style.height = "300px";

}

br.extend(C3Component, Component);

C3Component.prototype.getElement = function(){
	return this._rootElement

}

C3Component.prototype.setDisplayFrame = function(frame){
	this.m_oFrame = frame;
	frame.setContent(this.getElement());
	
}

C3Component.prototype.attachChart = function(){
	
	var chart = c3.generate({
		bindto: this.getElement(),
		data: {
			columns: [
				['data1', 30, 200, 100, 400, 150, 250],
				['data2', 50, 20, 10, 40, 15, 25]
			]
		}
[...]

Issue #1373, opened 2015-05-08
Elements not fully initialized at binding time


oniseijin:
The width and height of a DIV is unavailable at initial binding time. This may be a timing issue with how bindings work, but it makes the following code not work within an init:

d3.select(element).style("width") -> NAN, does not pick up style
d3.select(element).node().getBoundingClientRect().width  -> this is 0, but should be set to what the css sets it to, or what the page has/will render. Perhaps this is purely a drawback of Knockoutjs.  

Based off an attempt to reuse the bindings at https://github.com/teodoran/d3-knockout-demo

Code snippit
```
ko.bindingHandlers.barChart = {
	init: function(element, valueAccessor, allBindings, data, context) {
		"use strict";
         var elementWidth = d3.select(element).node().getBoundingClientRect().width // will be 0
         var elementWidth = parseInt(d3.select(element).style("width"), 10) // will be NAN
        }
}
```

Template
```
<div id="kendoui.d3bar.view-template">
	<div class="row">
		<div class="col-md-12 bigChart" data-bind="barChart: barChartData, style: {width: '500px' , height: '500px' }"></div>
	</div>
</div>
<!-- bigChart sets the height/the class col-md-12 setting the width, but being completely ignored -->
```

So, I had to hack using some arbitrary observable value, but this is an untenable work around

dchambers:
Ah, interesting, like #1372 this will also be caused by the template element not being added to the page at the the point we invoke `ko.applyBindings()`, but this time the code isn't doing anything unreasonable (like adding click listeners to global page elements). For components themselves there is a separate call-back when the element becomes attached to the page, but for knockout plug-ins we are initializing these before that event.

Let me take a proper look at this, and see what can be done...

oniseijin:
@dchambers Thanks: #1374 may be more of the same (and the most important, as I think the C3 library, based off D3 would be an extremely valuable addition). It is likely suffering from the ordering of initialization.

andyberry88:
This is fixed by #1379 which will be in the BRJS 1.0 release.

Issue #1372, opened 2015-05-08
Better support for Jquery plugins within ko bindingHandlers


oniseijin:
Something prevents EasyTree JQuery plugin from functioning fully. 

Problem: The knockoutjs binding, on the surface works, but is completely non-interactive. This is true for the work space version as well. 

Static export done here: http://signaller-rat-84376.bitballoon.com/
-The top tree menu is done using a knockoutjs binding [NON-INTERACTIVE]
-The bottom using a standard binding to a div on the main index.html [WORKING] (although, this approach does not work on the workbench because of div id creation/timing) workbench]
-workbench is also non-interactive with ko binding

So
A) why is the menu non-interactive within Bladerunner's standard framework
B) why is the "div" on the workbench, when set directly on the index.html of the workbench also not working (blank)

Using http://www.easyjstree.com/
HTML
```
<div id="kendoui.easytree.view-template">
		<div data-bind = "easyTree: treeMenuData"></div>
</div>
```

Binding
```
ko.bindingHandlers.easyTree = {
	init: function(element, valueAccessor, allBindings, data, context){
		var easytree = $(element).easytree({data: ko.unwrap(valueAccessor())})
		context.easytree = easytree

		console.log("easytree:" + JSON.stringify(context.easytree))
	}

}
 
```

this.treeMenuData is within the ViewModel as:
```
this.treeMenuData = ko.observableArray(
		[
			{
				"href":"http:\/\/www.easyjstree.com",
				"hrefTarget":"_blank",
				"text":"Home"
			},
			{
				"children":[
					{
						"href":"http:\/\/www.google.com",
						"hrefTarget":"_blank",
						"text":"Go to Google.com"
					},
					{
						"href":"http:\/\/www.yahoo.com",
						"hrefTarget":"_blank",
						"text":"Go to Yahoo.com"
					}
				],
				"isActive":false,
				"isExpanded":true,
				"isFolder":true,
				"text":"Folder 1",
				"tooltip":"Bookmarks"
			},
			{
				"children":[
					{
						"text":"Sub Node 1"
					},
					{
						"text":"Sub Node 2"
					},
					{
						"text":"Sub Node 3"
					}
				],
				"isActive":false,
				"text":"Node 1"
			},
			{
				"text":"Node 2"
			}
		]
	)
``` 

Work Bench Screen:
[...]

PR #1379, opened 2015-05-11
Initialize knockout plug-ins after template added to page

dchambers:
This pull-request hopefully fixes these issues:

  1. #1372
  1. #1373
  1. #1374 

#1372 can be tested by using the workbench page from the `easytree` app within the [easytree-brjs-spike](https://github.com/dchambers/-easytree-brjs-spike) &mdash; without this pull-request it's not possible to expand or collapse the nodes within the tree.

**Note:** Except for the fact that CT are using their own version of `PresenterComponent`, there would otherwise be a high risk this change introduces backwards compatibility issues. Even so, there may be issues even in terms of BRJS only infrastructure.


dchambers:
@oniseijin, sorry, only just seen this comment, but I agree that we have a collateral problem at present, and are steering people down the wrong path. I've created #1383 to address this.

ioanalianabalas:
Dev reviewed - all good :+1:

PR #1382, opened 2015-05-12
minor changes needed to make the appcache work with 1.0

andyberry88:
- adds a spec test method
- fixes the version and timestamp when `serve` is run rather than recalculating the timestamp each time the version is requested as it breaks the browser's appcache mechanism if the version changes


ioanalianabalas:
Dev-reviewed :+1:

Issue #1354, opened 2015-04-21
BRJS todo app not loading in IE8


thanhc:
The following error is being thrown in IE8 when opening the todo app:

Unexpected call to method or property access, when calling the following line in app.js line 46:

todoAppEl.appendChild(m_eElement)

both todoAppEl and m_eElement are objects.

ioanalianabalas:
Fixed by #1385, plus a minor change to the brjstodo app.

andyberry88:
@thanhc can you add a comment explaining why this was moved back to dev?

ioanalianabalas:
@andyberry88, @thanhc told me about it and he is encountering this issue during testing (it's in the PR #1385):

This fix is causing any new app to be created to display blade workbench components in the app itself.

Steps:
create a new app and add a blade then launch app

ioanalianabalas:
I have left the following comment on the corresponding PR but thought I should share it here as well:

I have tested the scenario mentioned by @thanhc but it seems like this behaviour was not introduced by this branch. I have tried creating a new app on both this branch and the `develop` branch, and I am still seeing the blade in the app itself. @thanhc will be raising a separate issue for that.

thanhc:
The app is now loaded correctly in IE8

PR #1385, opened 2015-05-12
Manually creating elements for each HTML5 tags.

ioanalianabalas:
Fixes issue #1354.

The HTML5 tags needed to be created separately as elements for IE8. In addition to this, I have changed the brjstodo app to create the `section` after the js bundle, so that the shim would run before the element is created.


thanhc:
This fix is causing any new app to be created to display blade workbench components in the app itself.

Steps:
create a new app and add a blade then launch app

ioanalianabalas:
I have tested the scenario mentioned by @thanhc but it seems like this behaviour was not introduced by this branch. I have tried creating a new app on both this branch and the `develop` branch, and I am still seeing the blade in the app itself. @thanhc will be raising a separate issue for that.

PR #1388, opened 2015-05-13
Renamspacing fixes

andyberry88:
- renamspace references from `unbundled-resources` to Blade classes
- renamspace references from a Blade to it's parent Bladeset
- renamspace references from one Blade to another Blade class - this isn't something that's 'supported' but unfortunately due to a bug it's worked previously so we need to fix it for backwards compatability


ioanalianabalas:
I made some minor suggestions here and there as comments on the code. :)

andyberry88:
Thanks @ioanalianabalas, all good feedback. Changes made in the latest commit.

Issue #1391, opened 2015-05-14
IIFE warning emitted for non CommonJS code


dchambers:
The warning `The CommonJS test 'my-test' is not wrapped within an IIFE, which may cause unreliability in tests.` seems to be emitted even for tests that aren't CommonJS.

PR #1393, opened 2015-05-14
IIFE warning fixes

andyberry88:
PR for #1391
- log the IIFEE warning for nested JS tests
- only log the warning if the JS style is `common-js`


andyberry88:
Should have been raised against `develop` not `master`.

PR #1395, opened 2015-05-15
Fix for Verifier parsing values with new line characters

charliehw:
`GwtTestRunner.prototype._parseStatement` should perform a global token replace when it removes/reinstates new line characters. Otherwise it only supports values with a single new line.


thanhc:
Tested and verified, just needs dev review then ready to merge

andyberry88:
:+1: looks ok

Issue #1392, opened 2015-05-14
Newly imported apps no longer show in the dashboard


dchambers:
This is a regression, as you now have to restart the server before they show in the dashboard.

PR #1396, opened 2015-05-18
FileObserver fixes

andyberry88:
- PR for #1392 to fix new apps not being shown in the dashboard
- Adding a single suite of tests for both `FileObserver` implementations


ioanalianabalas:
Looks good :+1: @andyberry88, would you like us to have a team meeting to discuss the issues mentioned in your TODOs for some of the tests?

andyberry88:
> would you like us to have a team meeting to discuss the issues mentioned in your TODOs for some of the tests?

I don't think we need a meeting to discuss it, it's something that needs some investigation at some point. Given we've got additional test coverage compared to what we had before I'd suggest we put this down to tech debt and investigate if/when there's time. This particular area is also covered indirectly by other tests, it's just the Java File Watcher API that's particularly difficult to test with.

PR #1403, opened 2015-05-26
Changed the safari test-runner command to launch safari

SplinterHead:
The command to launch safari is incorrect. Current methods do not allow users to call safari directly. To open the native install of Safari, the open command is needed


ioanalianabalas:
Dev-reviewed :+1:

Issue #1384, opened 2015-05-12
Can't run tests if project is in a symlinked path


vinhlam:
When I try to run `brjs test` from a path that contains a symlink, I get the following exception

    Testing ....../tests/test-unit (UTs):
    Unexpected Exception:
    java.lang.RuntimeException: Unable to find test pack which represents the path /...actual_full_path...../tests/test-unit/js-test-driver
    	at     com.caplin.cutlass.command.test.testrunner.JsTestDriverBundleCreator.createRequiredBundles(JsTestDriverBundleCreator.java:50)

eg.

The actual path is  `~/Documents/development/myapp`

I have a symlink in `~` created using `ln -s ~/Documents/development/myapp`
and running  `brjs test` in `~/myapp`

ioanalianabalas:
Hi @vinhlam,

I am closing this issue because it appears to be a duplicate of #1342.

ioanalianabalas:
Based on the recent comment received in the other issue, reopening this because it does not seem to be a duplicate after all.

ioanalianabalas:
Fixed by #1405.

thanhc:
This appears to already have been fixed by #1386 so closing this ticket

andyberry88:
I'm still getting this when running tests in CT using BRJS 1.0.1.

Issue #1386, opened 2015-05-13
apps should be created inside the working dir


james-s-turner:
OSX 10.9.4
Steps to reproduce:
1. Install BRJS in directory $BRJS_HOME
2. Create directory $APP_HOME and open command prompt 
3. Execute command $BRJS_HOME/sdk/brjs create-app wibble
4. App is created in $BRJS_HOME/sdk/brjs-apps/wibble

Should be created in directory I ran command from.

andyberry88:
> Should be created in directory I can command from.

This is different from the behaviour previously agreed - although it does make sense that this should be the behaviour.

To avoid any confusion what we previously decided on was:
1. if we're inside an app directory then apps live inside the parent directory of that app
2. if we can't find an app traverse up the directory tree to find `brjs-apps`
3. if we can't find a parent `brjs-apps` then use `brjs-apps` that live next to the `sdk`.

To support `Should be created in directory I can command from.` we would need to change `#3` so that we fall back to the current working directory as the location for apps.

james-s-turner:
Two points:
1. Even according to rule #3 the app is created in the wrong place. It should be inside a brjs-apps directory beside the sdk directory NOT inside it.
2. I take the point about the original decision. It would be nice if we could do what was stated. Any idea how much effort it would be.

andyberry88:
> Even according to rule #3 the app is created in the wrong place. It should be inside a brjs-apps directory "next to" the sdk directory NOT under it.

True, missed that point. That's definitely not what we were after.

> I take the point about the original decision. It would be nice if we could do what was stated. Any idea how much effort it would be.

I agree your suggestion makes more sense, I just wanted to clarify what the current expected behaviour was to avoid confusion. It should be reasonably trivial to make the change. I'd suggest we do it before 1.0.

Issue #1342, opened 2015-04-15
Can't run tests from the SDK folder


janhancic:
I've upgraded our app to CT v4.2.0 which uses BRJS v0.15.4.0. After the upgrade I can't run the tests from the SDK folder any more. I can run them one time, and every time after that BRJS errors out on every blade/bladeset/lib that has tests with this exception:

```
Testing myapp\users-bladeset\blades\clientsearch\tests\test-unit (UTs):
Unexpected Exception:
java.lang.RuntimeException: Unable to find test pack which represents the path D:\solutions\myapp\apps\myapp\users-bladeset\blades\clientsearch\tests\test-unit\js-test-driver
        at com.caplin.cutlass.command.test.testrunner.JsTestDriverBundleCreator.createRequiredBundles(JsTestDriverBundleCreator.java:50)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runTest(TestRunner.java:370)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runTestAndRecordDuration(TestRunner.java:357)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runAllTestsInDirectory(TestRunner.java:328)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runAllTestsInDirectory(TestRunner.java:333)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runTests(TestRunner.java:164)
        at com.caplin.cutlass.command.test.testrunner.TestRunnerController.run(TestRunnerController.java:118)
        at com.caplin.cutlass.command.test.TestCommand.doCommand(TestCommand.java:55)
        at org.bladerunnerjs.utility.CommandRunner.run(CommandRunner.java:25)
        at org.bladerunnerjs.model.BRJS.runCommand(BRJS.java:373)
        at org.bladerunnerjs.utility.UserCommandRunner.doRunCommand(UserCommandRunner.java:34)
        at org.bladerunnerjs.utility.UserCommandRunner.run(UserCommandRunner.java:26)
        at org.bladerunnerjs.model.BRJS.runUserCommand(BRJS.java:378)
        at org.bladerunnerjs.runner.CommandRunner.run(CommandRunner.java:111)
        at org.bladerunnerjs.runner.CommandRunner.main(CommandRunner.java:63)
```

If I run the tests from outside the SDK folder (in the root of the app or in the blade) then everything works as expected.

[...]

PR #1405, opened 2015-05-28
Exception when app not in known app location 1384

ioanalianabalas:
Fixes issues #1384 and #1342.

An exception is now thrown if the app you are trying to test is not in what BRJS knows to be the apps folder location.

Moreover, log lines will now show the BRJS apps folder and the apps found within.

**This is dependent on #1386**


andyberry88:
Looks good @ioanalianabalas. Only one minor comment which I've made inline. It might also be worth adding an info log for the location of BRJS itself like you did for the app location. There's also some merge conflicts to fix before it goes into `test`.

ioanalianabalas:
@andyberry88, thank you for your feedback. I have now merged in develop, removed the dead code and added an info log for the BRJS location. Would you mind having a look at these, please?

andyberry88:
:+1: looks good @ioanalianabalas

thanhc:
@ioanalianabalas can you fix the merge issues, also I've tried to reproduce the two issues this is suppose to fix,  #1384 and #1342. But on develop which already has #1386 I can't reproduce the issues, if I checkout an earlier build then I see the issues, so I'm wondering what this extra pull request is for?

ioanalianabalas:
@thanhc the merge issues have now been fixed.

thanhc:
This isn't quite working, if I run brjs test ../brjs-apps/myapp from the sdk/ then I get an error saying its not an app, but if I run it from the directory above brjs test brjs-apps/myapp then it works, so it seems like its not treating the ../ correctly.

ioanalianabalas:
@thanhc this is now fixed.

PR #1406, opened 2015-05-28
Make generated a root file registry dir

### LINKED ISSUE #1400: FXMotif first load taking a long time
While trying to upgrade the FX Motif to use BRJS 1.0, the initial load of the Motif is taking a very long time 40+ seconds it even causes chrome to stop working and pop up the kill chrome now window, looking at the network tab it seems to take a long time to get an image 40 seconds, once this image is returned then the rest of the app loads fine.

The image path is http://localhost:7070/fxtrader/v/dev/cssresource/aspect_default/theme_cotton/images/logo_with_bg.png

There is a branch in the motif which has been upgraded to 1.0 called brjs1.0Upgrade

andyberry88:
Make the generated folder a root folder for the file modification registry so that any changes in `generated` directory don't propagate all the way to the to `sdk` directory and invalidate bundle set memoized values. Fixes #1400

**This contains #1396 so should be merged after that**


ioanalianabalas:
Dev-reviewed :+1:

PR #1407, opened 2015-06-01
Fixed a bug for TooltipControl.js where the tooltip would not render correctly

dogoku:
It is expected behaviour that if the element that the tooltip needs to attach itself to is not visible, the tooltip will not render at all.

But if the tooltip needs to visible immediately when the page loads and the parent element is not visible yet, the tooltip will not render.

For that reason, we need to delay execution until rendering is complete

This is easy to test in CT. Please talk to me for a url


andyberry88:
Hey @dogoku, thanks for the PR. Could you also write a test to go with the change? Thanks!

dchambers:
@andyberry88, apparently this change can only be seen in CT, and is not easy to replicate in BRJS, so I was inclined to allow it through provided it could be manually tested with CT. However, I've just looked at the code and seen that the _fix_ is to use a `setTimeout()`, which just means that we are swapping an easy to reproduce bug for a hard to reproduce bug, whereas I'd far prefer to have an easy to reproduce bug.

@dogoku, can't your code wait on some underlying event, and in fact wouldn't that then make it more testable?

dchambers:
I'm pretty sure this has already been fixed by #1378.

Issue #1408, opened 2015-06-02
apps folder must not become legacy behaviour


janhancic:
I'm test upgrading our app to CT that uses BRJS RC1.

I get this warning in the console:

```
BRJS now uses a folder named 'brjs-apps' for the location of your apps but the directory 'd:\solutions\st\sdk' contains both 'brjs-apps' and 'apps' folders. 'd:\solutions\st\sdk/apps' will be used for the location of apps but this legacy behaviour may be removed so you should move all existing apps into the 'd:\solutions\st\sdk/brjs-apps' directory.
```

First issue is that the paths reported are wrong. Both folders are in the root of the repository, not in the SDK folder.

The second issue is that making `apps` legacy is unacceptable. We will not move our apps to a different folder as that will cause all sort of different problems.

andyberry88:
We decided to rename the directory when we looked at allowing apps to be separate from the SDK. It was done so we could be 100% sure that the directory was for brjs apps and not another directory that happened to be called `apps`. 

Now we're made a change to default to the working dir as the apps location (#1386) we probably don't need this `brjs-apps` change and we can revert back to use `apps` as the expected location for apps, otherwise default to the current working dir.

@dchambers @ioanalianabalas thoughts?

dchambers:
@andyberry88, if you are sure this is no longer necessary than of course it would be much better to only have one location rather than two, and given that we've only started using this new location in a _release-candidate_ than I'm happy to just stop supporting `brjs-apps` completely.

ioanalianabalas:
I think that given we are currently using the working directory, we could make that change, definitely. Will we now be checking the existence of an `apps` directory in the current working directory or parallel to the `sdk`? What is the order of our checks?

1. Look for `apps` in current directory
2. Look for `apps` next to the sdk

andyberry88:
I'd suggest we do the following:
- recursively traverse up the directory tree checking for `app.conf` or the `apps` directory (we already do this)
- if nothing is found default to the current working directory for the apps location (this is added in #1386)

The change would be to remove the additional check for `brjs-apps` and rename our of `brjs-apps` directory back to `apps`. The PR https://github.com/BladeRunnerJS/brjs-site/pull/97 wouldn't now be needed either.

dchambers:
Sounds good :+1:, @ioanalianabalas, are you happy with this?

ioanalianabalas:
Yes, should work fine :+1:

[...]

PR #1417, opened 2015-06-03
Modified apps folder search to look for apps instead of brjs-apps.

ioanalianabalas:
Fixes issue #1408.

I have changed the expected apps directory to `apps`, which must be located next to the `sdk`. Otherwise, the current working directory will be the apps location. I have also modified all our tests accordingly.


thanhc:
got two small issues with this ticket:

1. it seems like instead of checking for app.conf to see if something is an app or not, we also assume a folder is an app if there is a .html file in it.

2. From <brjs home>/brjs-sdk/sdk if I delete the apps/ folder from brjs-sdk/ and then run ./brjs serve, ideally what should happen is that it checks if there is an apps folder next to the SDK and if not then create one and use it, but instead it seems to check if there is an apps folder and because there isn't it then assumes the SDK folder is the apps folder and prints out all the warning messages saying brjs-sdk/sdk/docs is not an app .... before then creating an apps folder and using it.  Also when I do this I get a SAXParseException:  

FATAL@null line:-1 col:-1 : org.xml.sax.SAXParseException; Premature end of file.
Failed startup of context o.e.j.w.WebAppContext{/j2eeify-app,file:/D:/Development/brjs/brjs-sdk/sdk/j2eeify-app/}
org.xml.sax.SAXParseException; Premature end of file.
        at org.apache.xerces.util.ErrorHandlerWrapper.createSAXParseException(Unknown Source)
        at org.apache.xerces.util.ErrorHandlerWrapper.fatalError(Unknown Source)
        at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)

andyberry88:
> it seems like instead of checking for app.conf to see if something is an app or not, we also assume a folder is an app if there is a .html file in it.

Since the app checking code was written we've done this. You're right though we should probably only consider something an app if `app.conf` is present. I think there were problems doing that though, I remember some tests failing when there was a stricter check. It would be worth making it stricter and seeing what fails.

> From /brjs-sdk/sdk if I delete the apps/ folder from brjs-sdk/ and then run ./brjs serve, ideally what should happen is that it checks if there is an apps folder next to the SDK and if not then create one and use it, but instead it seems to check if there is an apps folder and because there isn't it then assumes the SDK folder is the apps folder

IMO this is the correct behaviour and the creation of the apps dir should be the responsibility of another part of the code. Whenever a BRJS is created it is populated using the templates, this is what creates the `apps` dir. It would be wrong for the app dir detection code to also be creating directories.  This is something we should write a test for though.

ioanalianabalas:
Fixed issues and merged the `develop` branch in.

andyberry88:
I've also added to this PR a change so the discovered app log is only printed the first time apps are used since it looked like this was causing problems with some observer plugins.

andyberry88:
I've also fixed the bug where the `apps` directory was created inside of `sdk` if run via the `./brjs` command. @thanhc this is ready to be QA'd and merged now.

PR #1424, opened 2015-06-11
Removed caplin and caplinx from reserved namespaces and added br inst…

### LINKED ISSUE #1404: Remove 'caplin' and 'caplinx' from reserved namespaces
We need to remove the reserved namespaces from `org.bladerunnerjs.utility.NameValidator` and the dashboard so that Caplin users of BRJS can create apps with those namespaces. Originally raised by @hobojed.

ioanalianabalas:
Fixes #1404.

Removed `caplin` and `caplinx` from our reserved namespaces, added `br` instead and modified tests accordingly.

### DIFF
--- a/brjs-core-tests/src/test/java/org/bladerunnerjs/spec/app/AppTest.java
+++ b/brjs-core-tests/src/test/java/org/bladerunnerjs/spec/app/AppTest.java
@@ -114,8 +114,8 @@ public class AppTest extends SpecTest {
 	@Test
 	public void usingReservedKeywordAsAppNameSpaceThrowsException() throws Exception {
 		given(appTemplate).containsFile("some-file.blah");
-		when(app).populate("caplinx", "default");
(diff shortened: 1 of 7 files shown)


andyberry88:
@ioanalianabalas there are also some reserved namespaces in the dashboard that should be replaced. Since you're looking at something else I can take a look at this to move the `1.0` release forward.

ioanalianabalas:
Thank you for the help @andyberry88 :)

ioanalianabalas:
@andyberry88 I see you added both `br` and `brjs` as reserved namespaces in the dashboard. I only added `br` to the Java code. Should I add `brjs` as well?

andyberry88:
> @andyberry88 I see you added both br and brjs as reserved namespaces in the dashboard. I only added br to the Java code. Should I add brjs as well?

I'd suggest we should probably reserve both `br` and `brjs`. @dchambers thoughts?

Issue #1426, opened 2015-06-15
Bundle file writing improvements


dchambers:
We recently identified a number of opportunistic improvements that should be easy to make, and which it would be good to get into the hands of users early in case they fix the test reliability problems people are seeing. These are issues 11, 12 & 14 from [this comment](https://github.com/BladeRunnerJS/brjs/issues/1370#issuecomment-111986192).

andyberry88:
Fixed by #1427.

PR #1427, opened 2015-06-15
Bundle file writing improvements

andyberry88:
Fixes for #1426
- the bundle file delete code will now longer fail silently
- increasing the retry sleep time to 100ms so we try for longer
- using `bundleFile.isFile()` to check if the file has been created rather than incorrectly using the return code from `createNewFile()`


dchambers:
Excellent work @andyberry88 :+1: One minor question/point raised,

Issue #1381, opened 2015-05-12
support favicon.ico in apps


andyberry88:
`favicon.ico` in the root of apps isn't served or part of built apps at the moment. Create a content plugin to serve `favicion.ico` files in the root of the app folder.

ioanalianabalas:
Fixed by #1430.

thecapdan:
@ioanalianabalas is the favicon.ico meant to be in the dashboard apps root folder, or should it be in default-aspect? also, the link to the favicon.ico in the dashboards index.html is incorrect, it points to unbundled resources

PR #1430, opened 2015-06-17
Favicon ico 1381

ioanalianabalas:
Fixes issue #1381.

I have added a favicon.ico content plugin, for when the favicon is located inside the default aspect.


andyberry88:
There were a few changes to be made to this PR which i've made to save you switching from what you're currently working on. We can go through the changes later when you have a sec.

Issue #1302, opened 2015-03-25
BRJS could be more friendly when leftover test folders are present


janhancic:
Let's say I have a library (or a blade, bladeset, ...) that uses the "old" folder structure for tests:

```
tests/
  test-unit/
    js-test-driver/
      ...
```

I then decide to use the new simplified structure and move my tests to the new structure:

```
test-unit/
  ...
```

I do this with git and commit everything and everything is fine.

But then, another developer pulls my changes (it also often happens on build machines). Because Git doesn't clean down the folders properly (because there are the `bundles` folder in the old location which aren't tracked by Git), BRJS then fails to run the tests for that library, which is fair enough, it expects one location for tests.

But it errors out with this exception:

```
Testing fooapp\libs\foo-lib\test-unit (UTs):
Unexpected Exception:
java.lang.ClassCastException: org.bladerunnerjs.model.AppJsLib cannot be cast to org.bladerunnerjs.model.TestPack
        at com.caplin.cutlass.command.test.testrunner.JsTestDriverBundleCreator.createRequiredBundles(JsTestDriverBundleCreator.java:43)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runTest(TestRunner.java:368)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runTestAndRecordDuration(TestRunner.java:354)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runAllTestsInDirectory(TestRunner.java:325)
        at com.caplin.cutlass.command.test.testrunner.TestRunner.runTests(TestRunner.java:160)
        at com.caplin.cutlass.command.test.testrunner.TestRunnerController.run(TestRunnerController.java:118)
        at com.caplin.cutlass.command.test.TestCommand.doCommand(TestCommand.java:55)
        at org.bladerunnerjs.utility.CommandRunner.run(CommandRunner.java:21)
        at org.bladerunnerjs.model.BRJS.runCommand(BRJS.java:354)
        at org.bladerunnerjs.utility.UserCommandRunner.doRunCommand(UserCommandRunner.java:35)
        at org.bladerunnerjs.utility.UserCommandRunner.run(UserCommandRunner.java:27)
        at org.bladerunnerjs.model.BRJS.runUserCommand(BRJS.java:359)
        at org.bladerunnerjs.runner.CommandRunner.run(CommandRunner.java:104)
[...]

PR #1436, opened 2015-06-18
Added RuntimeException if there is an old test folder parallel to tes…

ioanalianabalas:
…t-unit or test-acceptance.

I have added a RuntimeException that is being through if there is a folder called `tests` next to the folder containing the test suite we are currently running.

Fixes issue #1302.


andyberry88:
@ioanalianabalas and I have spoken about a better way to do this which I'll take a look at so she can take a look at some JS issues.

PR #1447, opened 2015-06-29
Remove parser from editable property

andyberry88:
- allowing parsers to be removed from EditableProperty
- similar to removeValidator - returns true if a parser was removed, false otherwise.


andyberry88:
This should have been raised to merge into `develop` rather than `master`.

Issue #1453, opened 2015-07-02
Warning about strict mode in IE*


andyberry88:
The locale forwarding page doesn't have a `DOCTYPE` so IE doesn't load in strict mode and BRJS shows an error. Going directly to a locale page works.

Seen using BRJS 1.0. There's no workaround to make the redirection page work - the only way to load the app is to go directly to the specific locale URL.

![upload](https://cloud.githubusercontent.com/assets/726896/8479704/5565fe04-20d1-11e5-971e-8e3196f54ade.png)

andyberry88:
Fixed by #1454.

thecapdan:
tested with the IT-app in IE9 - works now

PR #1454, opened 2015-07-02
Fixing locale forwarder doctype

andyberry88:
PR for #1453.


andyberry88:
Closing to raise against `develop` rather than `master`.

PR #1460, opened 2015-07-09
An empty string should be returned if the JNDI lookup returns null.

jsorbello:
If a JNDI entry is set to an empty string the JNDI service returns null which prevents the use of empty values.


andyberry88:
:+1: Good PR. Thanks @jsorbello.

Issue #1071, opened 2014-11-14
[feature] App level alias override


janhancic:
I would like to be able to define an alias override that would apply to everything in the app (aspects, workbenches & tests). With the option to further override (as you can now) on workbench/test level.

Right now I have to copy the same alias override (in `aliases.xml`) to a lot of different places in my app.

andyberry88:
This was largely complete when I picked up it. I've added tests for groups and scenarios. @thecapdan can you do your thing and do some exploratory testing?

thecapdan:
I think we need to update the docs to demonstrate how to use this new feature, maybe here: http://bladerunnerjs.org/docs/use/ioc/

thecapdan:
The spec tests for this are a little unclear - so we should tidy those up + I'll add an example in the it-app

thecapdan:
"With the option to further override (as you can now) on workbench/test level." I assume this means we have app level aliases that we would like to optionally override at the test level. did we decide how this was going to be done, if so, can we document it

thecapdan:
Moving back to dev.

1. needs acceptance criteria in the description
2. when i add an aliases.xml with an alias defined at the app level, i am unable to use that alias in App.js. (I can add this test case to the it-app)

andyberry88:
While looking at whether we had a test case for the issue @thecapdan described above I noticed
a. we don't have a test for this, so we need to add one for this scenario and any related scenarios
b. some of the tests are badly named: `aliasDefinitionsInAspectAreOverridenByTheAppAliases` isn't the behaviour we expect and what the test is actually testing. This should be something like `aliasDefinitionsInAspectOverrideAppAliases`.

This needs a lot more work.

thecapdan:
retested. working now. + overriding the app level alias in the default aspect works as expected.

needs release note, but will merge anyway because im a soft touch

PR #1486, opened 2015-07-31
App level alias override 1071

andyberry88:
PR for app level aliases (https://github.com/BladeRunnerJS/brjs/issues/1071).


plugins/brjs-plugins/src/test/java/org/bladerunnerjs/spec/aliasing/AppLevelAliasOverrideTest.java
```diff
@@ -0,0 +1,382 @@
...
+		bladeAliasDefinitionsFileBuilder = new AliasDefinitionsFileBuilder(this, aliasDefinitionsFile(blade, "src"));
+		bladeTestPackAliasesFileBuilder = new AliasesFileBuilder(this, aliasesFile(bladeTestPack));
+		
+		servicesLib = brjs.sdkLib("ServicesLib");
+		given(servicesLib).containsFileWithContents("br-lib.conf", "requirePrefix: br")
+			.and(servicesLib).hasClasses("br/AliasRegistry", "br/ServiceRegistry");
+	}
+	
+	@Test
+	public void aliasDefinitionsInLibsAreOverridenByTheAppAliases() throws Exception {
```
dchambers: These tests are psuedo BDD since `brLibAliasDefinitionsFileBuilder` and `appAliasesFileBuilder` contain technical information about the underlying type &mdash; that it's a builder. We should update all tests to remove anything like this.

andyberry88: Not 100% sure what you mean by this. Whatever we do, since Aliases are a plugin and not part of core BRJS, we can't add this to the spec test infrastructure like we've done for creating apps etc.

dchambers: I mean to rename `appAliasesFileBuilder` to `appAliasesFile`, so that it's written in BDD style. The fact the thing uses the _builder_ pattern should not leak into our spec tests.

Issue #1494, opened 2015-08-11
build-app throws exception when using a version containing underscores


thecapdan:
to reproduce, build an app called bob with the following comman:

```
brjs build-app bob -v 1.2.3_4 
```

PR #1495, opened 2015-08-12
Allow underscores in versions 1494

thecapdan:
fixes https://github.com/BladeRunnerJS/brjs/issues/1494


andyberry88:
@thecapdan The changes in this PR look good but it looks like there are some genuine failures in Travis.

PR #1508, opened 2015-08-28
fix: fix `brjs` start script so it works for latest cygwin

briandipalma:
The newest version of `git` for windows seems to be using a different `uname`
value.

### DIFF
--- a/brjs-sdk/sdk/brjs
+++ b/brjs-sdk/sdk/brjs
@@ -21,7 +21,7 @@ fi
 SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
 BRJS_CLASSPATH="$SCRIPT_DIR/libs/java/system/*:$SCRIPT_DIR/../conf/java/*"
 
-if [ "$(echo $(uname -s) | cut -c 1-6)" = "CYGWIN" ]; then
+if [[ "$(echo $(uname -s) | cut -c 1-6)" = "CYGWIN" || "$(echo $(uname -s) | cut -c 1-6)" = "MINGW6" ]]; then
 	# Cygwin version
 	java $JAVA_OPTS -cp "`cygpath -wp $BRJS_CLASSPATH`" org.bladerunnerjs.runner.CommandRunner "`cygpath -wp $SCRIPT_DIR`" "$@"
 else


dchambers:
Thanks @briandipalma, good find and thanks for the pull-request. We'll get @thecapdan to verify this on Tuesday since it seems weird that Cygwin is starting to report `MinGW` since that's what GitBash uses, and I understood to be more low-level as it doesn't support POSIX.

thecapdan:
Hi @briandipalma - can you clarify your comment, did you mean to say 'git' in your description or is this just about cygwin?

If we're to believe the [wikipedia page on uname values](https://en.wikipedia.org/wiki/Uname) then it should still give us the correct value.

Also, I have tried this on what I assume is the latest version of cygwin and the unix name is still "Cygwin". Are you using some kind of tabbed environment like conemu?

briandipalma:
Try

>newest version of git for windows

https://git-for-windows.github.io/

it
> seems to be using a different uname value.

Maybe it's just my machine but I don't think I have anything weird installed on it...

thecapdan:
thanks for clearing that up @briandipalma - I think @dchambers and I were confused by the name of the PR suggested the issue was with cygwin.

As far as i can tell, when using gitbash, uname -s has always returned MINGW. But the latest version requires cygpath conversion.

I downloaded the latest git and I had the same issue as you're seeing. Your solution works but it would not be compatible with older versions of git, those using the older versions would then get this error:

```
./brjs: line 26: cygpath: command not found
```

Also, I believe your solution assumes the user will be using the 64 bit version. If they have the 32 bit version then echo $(uname -s) | cut -c 1-6) returns MINGW3, But it doesn't look like this check is what's needed anyway.

Maybe @andyberry88 has a suggestion

briandipalma:
```bash
if ! type cygpath; then
	....
fi
```

thecapdan:
'type cygpath' also throws an error in the old gitbash.

The following line works:

```
if [[ "$(echo $(uname -s) | cut -c 1-6)" = "CYGWIN" || "$(uname -a)" = *"Msys"* ]]; then
```

This works for all versions of git and cygwin

i would have used the -o flag (shows the OS), which would have looked tidier but the old bash console doesn't support this flag.

briandipalma:
`type` is a test, so yes you will get an error logged but the script should continue.

That's why you pipe to `/dev/null`.

thecapdan:
closing in favour of https://github.com/BladeRunnerJS/brjs/pull/1510

thecapdan:
sorry @briandipalma didn't mean to cut you off. you are right, when I tried 

```
if ! type cygpath > /dev/null; then
```

I got an error, but removing the NOT worked, but with a warning. I've opened another PR with my proposed solution, but yours does seem tidier if we can suppress the warning

andyberry88:
Re-opening this since the approach in #1510 could potentially cause problems if a `cygpath` command is on their path - which *could* happen if paths are shared across different CLI environments.

IMO we should go for the `uname` approach.

thecapdan:
closing in favour of https://github.com/BladeRunnerJS/brjs/pull/1518

dchambers:
This pull-request causes spurious console output on Ubuntu since:

  1. For compatibility reasons we use `#!/bin/sh` instead of `#!/bin/bash`.
  1. For performance reasons, `/bin/sh` points to `/bin/dash` ([Debian Almquist Shell](https://wiki.archlinux.org/index.php/Dash)) rather than `/bin/bash` on Ubuntu.
  1. Dash doesn't support the [newer double square bracket notation](http://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash#answer-52050).

This is what I now see:

```shell
./brjs serve
./brjs: 25: ./brjs: [[: not found
./brjs: 25: ./brjs: Linux dominicc-laptop 3.16.0-23-generic #31-Ubuntu SMP Tue Oct 21 17:56:17 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux: not found
```

@briandipalma, is there a particular reason you needed the double square bracket notation, or would it be okay for me to change it back to single bracket notation?

PR #1510, opened 2015-09-01
Fix: brjs start script does not work on latest version of gitbash

thecapdan:
found by @briandipalma, when downloading the latest version of git, the version of gitbash supplied needs to use the same command as cygwin.

As suggested by brian, the following also works, but we get a warning message in the old gitbash:

```
if type cygpath > /dev/null; then
```


thecapdan:
having spoken to @andyberry88 , he preferred @briandipalma 's [solution to mine](http://memeguy.com/photos/images/mrw-people-tell-me-to-stop-drinking-stop-staying-up-late-and-to-get-more-exercise-8867.gif). So I've checked in a modified version which doesn't print the output of the type to the console. should be ready to merge once we've ascertained the script works in cygwin

thecapdan:
works in cygwin, ready to merge if you're happy @andyberry88

thecapdan:
looks like travis doesn't like the latest solution. @andyberry88 as a mac user, you are in the best position to investigate the problem ;)

andyberry88:
Since this approach relies on `cygpath` being on the `PATH` it's potentially unreliable and can easily lead to false positives if `cygpath` happens to be available but Cygwin isn't the active CLI. Closing this and I've reopened https://github.com/BladeRunnerJS/brjs/pull/1508.

PR #1513, opened 2015-09-08
Make ViewFixture errors more useful

janhancic:
Given this line in a test:

```javascript
and("ticket.view.(.my-selector).text = 'XYZ'");
```

if no element was found for `.my-selector`, you would get this error message:

```
No view element found for "text"
```

After this change, you get this:

```
No view element found for "text". Processing property "(.my-selector).text" and looking for value "XYZ"
```

which is obviously much much better as it actually helps you pin point the location in the test where the problem is.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-test/src/br/test/ViewFixture.js
+++ b/brjs-sdk/sdk/libs/javascript/br-test/src/br/test/ViewFixture.js
@@ -267,7 +267,7 @@ ViewFixture.prototype._getHandler = function(propertyName, value) {
 		if (handler.elements.length === 1) {
 			handler.selectedElement = handler.elements[0];
 		} else {
-			this._verifyOnlyOneElementSelected(handler.elements, handler.property);
+			this._verifyOnlyOneElementSelected(handler.elements, handler.property, propertyName, value);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


sospirited:
Completely for this item to be merged, it's one of those things that you can't appreciate good error messages until you have to get to the bottom of these types of issues.

Should be straight forward to add some UT coverage for this item too.

janhancic:
Also worth pointing out that you don't get a call stack when this failure happens, so it's really really hard to figure out what failed.

dchambers:
Thanks @janhancic.

janhancic:
:+1:

thecapdan:
@janhancic _verifyOnlyOneElementSelected is also used in the setup with 2 arguments. If the ViewFixture is setup with the wrong selector we would now get the following error:

```
'No view element found for 'badSelector'. Processing property '' and looking for value ''
```

We should either check for the additional args you added or provide a nice message for the case where it fails during setup

janhancic:
I missed that. I'll submit a fix.

brjs-sdk/sdk/libs/javascript/br-test/src/br/test/ViewFixture.js
```diff
@@ -291,11 +291,13 @@ ViewFixture.prototype._getViewElements = function(propertyName) {
 };
 
 /** @private */
-ViewFixture.prototype._verifyOnlyOneElementSelected = function(elements, selector) {
+ViewFixture.prototype._verifyOnlyOneElementSelected = function(elements, selector, propertyName, value) {
 	if (elements.length === 0) {
-		throw 'No view element found for "' + selector + '"';
+		throw 'No view element found for "' + selector + '". Processing property "' + propertyName +
```
thecapdan: here 'selector' isn't a selector, its a view handler. Can you change the arg name to viewHandler

PR #1515, opened 2015-09-09
change HtmlResourceService to only loads HTML when no templates are loaded

andyberry88:
Performance improvement. Changes HTML resource service so it only loads templates once rather than every time the class is constructed which may be multiple times when tests are run.


brjs-sdk/sdk/libs/javascript/br-services/src/br/services/html/BRHtmlResourceService.js
```diff
@@ -2,6 +2,8 @@
 * @module br/services/html/BRHtmlResourceService
 */
 
+var TEMPLATES_LOADED = false;
```
dchambers: I think this should be renamed to `templatesLoaded` since the all upper case convention is used for constants rather than static variables.

andyberry88: Yep, true. Fixed.

thecapdan:
tests are running smoothly in IE9, seems to have fixed the performance issue

Issue #1506, opened 2015-08-26
Fix MappedNodeList broken paths bug


dchambers:
Start by creating a test case to show that adding a node to a `MappedNodeList` causes the paths to break for any nodes previously added to the `MappedNodeList`. Additionally, it seems like if the items within the mapped node-list have private references to nodes from other parts of the presentation model, these paths are broken too, and so we should add a test for this also.

dchambers:
@andyberry88, I've marked this issue as a `high-priority` since it's blocking work for two separate client teams. Would you be able to look at this as soon as you've finished what you're working on? Even if we only get a couple of failing tests than that would be really helpful.

andyberry88:
I've taken a look at this, here's what we know so far.

- The exec summary is: it looks like this isn't a bug in core BRJS or CT code so it's probably being caused by a custom JS patch.
- CT and BRJS have different implementations of `PresentationNode` which handles the path logic.
- In both BRJS and CT the paths before and after nodes are added to a MappedNodeList are correct.
- In CT nodes with the prefix `m_` or `_` are ignored when creating node paths
- In BRJS *no* nodes are ignored when creating paths. This means if a `MappedNodeList` item has a 'private' reference to another node in the presentation model and exception will be thrown since it deemed to be in the tree twice. This looks like a design design taken when splitting off the BRJS version of the Presenter codebase but seems wrong since it prevents internal references to other nodes.

To summarise: not a bug in our code so I'm going to remove the `high priority` label, but we should probably discuss whether or not BRJS should ignore `_` prefixed properties when it comes to path calculation (probably when @dchambers is back from holiday).

dchambers:
Yes, I think we should support '-' and 'm-' too, since we ultimately want to merge these two code bases back into one.

PR #1516, opened 2015-09-09
m_ and _ fields are now private and extra tests for MappedNodeList

andyberry88:
- PresentationNode fields prefixed `_` or `m_` are treated as private
- Added tests to ensure MappedNodeList doesn't break PresentationNode paths - #1506


brjs-sdk/sdk/libs/javascript/br-presenter/test-unit/src-test/br/presenter/testing/node/RootPresentationNode.js
```diff
@@ -6,6 +6,7 @@ var WritableProperty = require("br/presenter/property/WritableProperty");
 var RootPresentationNode = function()
 {
 	this.m_oPrivateProperty1 = new WritableProperty();
+	this._oAnotherPrivateProperty1 = new WritableProperty();
```
thecapdan: @andyberry88 you've added this new property but don't seem to be using it in any of the tests

andyberry88: It's used by the fact that it doesn't show up as a property in the tests that asset against `properties().length`.

thecapdan: ok, thanks for clarifying

PR #1518, opened 2015-09-10
ensure brjs command now works on latest gitbash versions

thecapdan:
latest gitbash version needs the cygwin style command, so additional check required for brjs commands


dchambers:
This pull-request causes spurious console output on Ubuntu since:

  1. For compatibility reasons we use `#!/bin/sh` instead of `#!/bin/bash`.
  1. For performance reasons, `/bin/sh` points to `/bin/dash` ([Debian Almquist Shell](https://wiki.archlinux.org/index.php/Dash)) rather than `/bin/bash` on Ubuntu.
  1. Dash doesn't support the [newer double square bracket notation](http://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash#answer-52050).

This is what I now see:

```shell
./brjs serve
./brjs: 25: ./brjs: [[: not found
./brjs: 25: ./brjs: Linux dominicc-laptop 3.16.0-23-generic #31-Ubuntu SMP Tue Oct 21 17:56:17 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux: not found
```

@briandipalma, is there a particular reason you needed the double square bracket notation, or would it be okay for me to change it back to single bracket notation?

briandipalma:
```bash
$ ./brjs serve                                                                                                        
./brjs: line 24: [: missing `]'                                                                                       
./brjs: line 24: MINGW64_NT-6.1 BRIAND-T420 2.1.0(0.288/5/3) 2015-06-28 18:57 x86_64 Msys: No such file or directory  
```

:/

Don't know what to suggest tbh I'm no bash expert.

briandipalma:
Who has compat issues with using `#!/bin/bash`?

dchambers:
Well it was originally done so we could support all UNIXes, including BSD and Solaris, but I'm not sure if anybody has ever used those platforms in practice. If there's a reason why you absolutely need double bracket syntax then we'll switch to requiring bash?

dchambers:
Sorry @briandipalma, just seen your earlier comment. Leave it to me...

dchambers:
The Ubuntu regression should be fixed by #1528.

PR #1524, opened 2015-09-15
Add new Closure minification setting

dchambers:
Added a new `closure-medium` minification level between `closure-simple` and `closure-advanced` that causes private members (i.e. members starting with `_` or `m_`) to be obfuscated. This change is dependent on the following up-stream improvement for Closure Compiler:

  * https://github.com/google/closure-compiler/pull/1129

I've so far been able to confirm that the 'it-app' app works with this new obfuscation level, but to progress this further somebody (@andyberry?) will need to confirm that the other more complex apps continue to work, and to find a way for us to start making use of the Closure Compiler changes pending a new release of Closure Compiler that contains those changes.

For the time being, you can test things by compiling [brjs-closure-compiler](https://github.com/BladeRunnerJS/closure-compiler/commits/rename-public-properties-compiler-option) on your own machine, and then updating the `closure-compiler` line in `dependencies.gradle` to something like this:

```
'closure-compiler': files('/home/dominicc/git-dev/brjs-closure-compiler/build/compiler.jar'),
```


dchambers:
Good work @andyberry88 :+1:

dchambers:
@Aivarass just mentioned the lack of release notes for this at present. @andyberry88, do you think we should keep this as a dark feature for the moment given the limited amount of automated testing we're able to perform? I'm wondering if we should wait until we get positive feedback from CT before announcing it?

andyberry88:
I was mentioning to @Aivarass we should probably have some form of release notes for this. Releasing the `closure-medium` part as a dark feature is probably a good idea but the `--js-minifier` flag for tests might be something that's worth advertising since it accepts all of the various minifiers that can be specified in the `js.bundle` tag and could be useful.

dchambers:
Yes, good point.

PR #1528, opened 2015-09-16
Don't print spurious characters on Ubuntu

dchambers:
Because `bin/sh` points to `bin/dash` rather than `bin/bash` on Ubuntu,
we can't use double square bracket notation, so use `grep` instead.


thecapdan:
i tried this in the latest gitbash on windows and i get the same issue as before:

```
$ ./brjs serve
```

PR #1529, opened 2015-09-17
Add destroy method for jQueryAutoCompleteControl to clean listeners

michalvankodev:
When using components which are added/removed through the use of the application which use this autocomplete-box controller they leak memory caused by not cleaned listener and not destroyed jQuery autocomplete widget.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/control/selectionfield/JQueryAutoCompleteControl.js
+++ b/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/control/selectionfield/JQueryAutoCompleteControl.js
@@ -11,8 +11,6 @@ var Core = require('br/Core');
  * @module br/presenter/control/selectionfield/JQueryAutoCompleteControl
  */
 
-var jQuery = require('jquery');
-
 /**
  * @class
  * @alias module:br/presenter/control/selectionfield/JQueryAutoCompleteControl
@@ -51,6 +49,7 @@ function JQueryAutoCompleteControl() {
 	this.m_eElement = {};
 	this.m_bOpenOnFocus = false;
 	this.m_sAppendTo = 'body';
+	this._viewOpened = false;
 }
 
 Core.inherit(JQueryAutoCompleteControl, ControlAdaptor);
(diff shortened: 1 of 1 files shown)


brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/control/selectionfield/JQueryAutoCompleteControl.js
```diff
@@ -11,8 +11,6 @@ var Core = require('br/Core');
  * @module br/presenter/control/selectionfield/JQueryAutoCompleteControl
  */
 
-var jQuery = require('jquery');
```
janhancic: Why did you remove this? It's a dependency.

michalvankodev: Its duplicated .. jQuery is defined as first dependency at the top

brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/control/selectionfield/JQueryAutoCompleteControl.js
```diff
@@ -163,4 +161,12 @@ JQueryAutoCompleteControl.prototype.setOptions = function(mOptions) {
 	}
 };
 
+/**
+ * Destroy created listeners and jQuery autocomplete plugin
+ */
+JQueryAutoCompleteControl.prototype.destroy = function() {
+	this.m_oPresentationNode.value.removeListener(this._valueChangedListener);
```
thecapdan: @michalvankodev can we add a gueard here in case detroy is called without any nodes being set... e.g. if(this._valueChangedListener). similarly a guard should be added for this.m_eElement on the line below

michalvankodev: Yes, Today we found out that not all Nodes that use jQueryAutoCompleteControl had triggered `onViewReady` and then it will break the execution because the `autocomplete` plugin will throw error as no autocomplete was set.

michalvankodev: I've updated the PR wth the checks.

PR #1537, opened 2015-09-24
add a ServiceRegistry.dispose() method

andyberry88:
add `ServiceRegistry.dispose()` which in turn calls `dispose()` on any services that have a dispose method. Ignores services without `dispose()` and those where `dispose` accepts more than 0 args.


brjs-sdk/sdk/libs/javascript/br/src/br/ServiceRegistryClass.js
```diff
@@ -129,6 +129,29 @@ ServiceRegistryClass.prototype.legacyClear = function() {
 	this.registry = {};
 };
 
+ServiceRegistryClass.prototype.dispose = function() {
```
dchambers: Is there a reason we are logging to the console instead of _fell_? If we should be using _fell_ than we should probably take the opportunity to log at more appropriate log levels as some of the information level log lines seem like they should really be debug level.

andyberry88: No reason, just an oversight. I'll change it to use `fell`.

dchambers: :+1:

PR #1538, opened 2015-09-24
change serve and build cmds so dev and prod are default environments

andyberry88:
also changes build command so it only appends the environment to the war
of dir name if the environment is specified via the '-e' flag


thecapdan:
@dchambers this is pretested, you can merge once its been code reviewed. thanks

dchambers:
@thecapdan, you may want to re-test this as I've made some changes because the help information either wasn't being displayed at all (in the case of `brjs serve`) or was repeating itself regarding the default `environment` value (in the case of `brjs build-app`).

thecapdan:
@dchambers build failure looks real:

```
getCommandHelp() in MultipleArgsTestCommand cannot override getCommandHelp() in JSAPArgsParsingCommandPlugin
	public String getCommandHelp() {
	              ^
  overridden method is final
1 error
:brjs-runner:compileTestJava FAILED
FAILURE: Build failed with an exception

```

PR #1543, opened 2015-09-25
Treat linux64 as a seperate browser 'OS'

dchambers:
By supporting 64-bit browsers on Linux, we are now able to run browser
tests (specifically Firefox) on Travis.


dchambers:
@andyberry88, I've switched Travis from running `:brjs-sdk:test` (PhantomJs) to `:brjs-sdk:testFirefox`, where I'm assuming that the only material difference between these two targets is the browser in use?

andyberry88:
@dchambers correct. I've dev reviewed this and all looks good. Probably something we need to highlight in the release notes since its potentially a change to which browser path we use on a devs machine.

PR #1545, opened 2015-09-29
v1.1.1 release notes added

thecapdan:


releases/v1.1.1.json
```diff
@@ -0,0 +1,5 @@
+{
+    "milestone": "1.1.1",
+    "name": "BladeRunnerJS 1.1.1",
+    "prerelease": "false"
```
andyberry88: @thecapdan This needs to be `false` rather than `"false"` - this is what was causing the error we saw last time we tried to release.

thecapdan: oops - thanks, clearly i chose the wrong one to copy paste from!

andyberry88:
:+1: Looks good now. This can be merged once Travis passes.

PR #1555, opened 2015-10-02
BRJS should now recognises and uses compiled directories

trueadm:
In order to work with Caplin's Mobile new workflow in using a compiler, a specific "compiled" directory is required to be used instead of the "src" directory. If no "compiled" directory is found, "src" should be used instead.


andyberry88:
I'm not sure this is something we want in core and should probably be a custom plugin.

trueadm:
I disagree there, I think this is something that is ideal for environments where teams are using third-party tools to handle JS compilation.

andyberry88:
IMO its a very specific use case (the clue being "In order to work with **Caplin's Mobile** new workflow"). It also requires that thirdparty tools are configured in a very specific way to output their compiled code in the correct format. It would be best either in a separate plugin hosted under the BRJS org or a completely separate plugin hosted elsewhere.

andyberry88:
Spoke to @trueadm about this. I'm going to give the team a skeleton Asset plugin they can change/use for their needs. I'm going to close this for the time being and we can re-open it if the plugin route doesn't work. This PR will need tests and some minor refactoring before it's merged.

PR #1568, opened 2015-10-22
fix bug where non default aspect index page was wrongly named when built

andyberry88:
We have a bug where `index.html` is named `.html` when building an app with a non default aspect and a single locale. This PR fixes that bug.


andyberry88:
**Before** this is merged we should establish whether we want to do a patch release on `1.1.1` - since this is a pretty major bug that basically prevents building single locale apps with multiple locales. If we do want a patch release this PR should be re-raised against a patch branch based from `1.1.1` and that branch then merged back in to master after the release. // @dchambers @thecapdan @Aivarass

dchambers:
Personally, my preference is to always commit to `master` first on the basis that nobody will ever forget to test a patch release that only contains that one feature, yet I have seen changes accidentally never make it back into `master`, or (very rare) be merged but not work properly.

Also, this will need to be tested in both releases separately (if there are to be two releases) since these two branches won't contain this commit in their common ancestry.

dchambers:
Oh, if people are shouting for this then it makes sense that we provide a release. If there have been any other risky changes in the interim then it also makes sense that we do a patch release containing only this change.

andyberry88:
The problem with committing to `master` and then cherry picking to a release branch is that the commit ID changes (since the ID is based on the previous commit). This can make debugging, or establishing whether a commit is in a given release, more difficult later since the 2 branches are entirely different.

I'd suggest that we merge this to a `1.1.2` patch release and then the `1.1.2` branch is merged in to master - the difference then between the two branches here should be 1 commit.

dchambers:
> The problem with committing to master and then cherry picking to a release branch is that the commit ID changes (since the ID is based on the previous commit). This can make debugging, or establishing whether a commit is in a given release, more difficult later since the 2 branches are entirely different.

That's only a problem if you cherry-pick to `master` though. If you cherry-pick from `master` then you can see every branch a commit is in except the patch release, which only exists to contain that one feature anyway, and so is obvious.

> I'd suggest that we merge this to a 1.1.2 patch release and then the 1.1.2 branch is merged in to master - the difference then between the two branches here should be 1 commit.

If that's your preference then lets do that :+1:.

andyberry88:
> That's only a problem if you cherry-pick to master though. If you cherry-pick from master then you can see every branch a commit is in except the patch release

True. Unless the patch release also needs a patch. :smile: 

> If that's your preference then lets do that

:+1: I'll sort out the release branch and the PRs.

andyberry88:
I've spoken to @sospirited and we've decided that this can be fixed in the next planned BRJS release.

thecapdan:
tested with a single locale app with a non default aspect. works + tests reviewed

PR #1570, opened 2015-10-29
Add getter function for Properties

michalvankodev:
I'd like to attach listeners to the properties of `Properties` via `PropertyHelper` to avoid memory leaks.
PropertyHelper accepts only `Property` objects as arguments.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/property/Properties.js
+++ b/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/property/Properties.js
@@ -60,6 +60,15 @@ Properties.prototype.add = function(vProperties) {
 };
 
 /**
+ * Returns array of properties in the collection.
+ *
+ * @type Array
+ */
+Properties.prototype.getProperties = function() {
+	return this.m_pProperties.slice(0);
+};
+
+/**
  * Returns the size of the collection.
  * 
  * @type int
--- a/brjs-sdk/sdk/libs/javascript/br-presenter/test-unit/tests/br/presenter/property/PropertiesTest.js
+++ b/brjs-sdk/sdk/libs/javascript/br-presenter/test-unit/tests/br/presenter/property/PropertiesTest.js
@@ -30,6 +30,19 @@
         
     };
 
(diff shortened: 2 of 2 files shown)


dchambers:
Hi @michalvankodev,

`Properties` already has an `addListener(listener)` method, where `listener` supports the following callbacks:

  * `onPropertyUpdated`
  * `onPropertyChanged`
  * `onValidationSuccess`
  * `onValidationError`
  * `onValidationComplete`

If these aren't sufficient for what you need could you please provide some more context so we can better understand your actual needs? Or, perhaps you can link the commit where you fix the memory leak you're seeing?

michalvankodev:
Hi @dchambers I am not fixing a memory leak. I'm avoiding creating one with usage of `PropertyHelper` which will track properties I attach listeners to. As I am using `SLJSContainerManager` I have no control of the properties which I attach listeners to as those can be removed at any time. So I attach listeners via `PropertyHelper` which allows me to remove listeners easily. I just need to get the instance of the `Property` and not `Properties` as `PropertyHelper` only accepts those.

dchambers:
Hi @michalvankodev, any chance you could point me to the code snippet in question so I can take a look for myself?

dchambers:
Sorry for the delay @michalvankodev, I've only just now had a chance to look at this. AFAICT, this is the relevant bit of code:

~~~js
XXXPM.prototype._setUpNodeListeners = function() {
	var nodes = this.rows.getPresentationNodesArray();

	this._propertyHelper.removeAllListeners();

	nodes.forEach(function(node) {
		var properties = node.properties().getProperties();

		properties.forEach(function(prop) {
			this._propertyHelper.addChangeListener(prop, this, '_onNodesChanged', false);
		}, this);
	}, this);
};
~~~

Excuse my ignorance, but what is `PropertyHelper` and why are we using it? It's not currently clear to me what it adds over just using `Property` directly.

dchambers:
@michalvankodev, could you please change your method to this:

~~~js
Properties.prototype.getProperties = function() {
	return this.m_pProperties.slice(0);
};
~~~

so that the original can't be mutated.

dchambers:
Thanks @michalvankodev :+1:

PR #1579, opened 2015-11-09
updating selenium version from 2.44.0 to 2.48.2

Aivarass:


andyberry88:
This was backed out in #1590 since it causes all sorts of issues with changes in transitive dependencies.

PR #1583, opened 2015-11-12
Removed the new line character at the end of the CommonJsSourceModule…

### LINKED ISSUE #1567: On bundled code, remove newline from the wrapping function
Currently, BRJS adds this line to the top of each file/module:

`define('module', function(require, exports, module) {`

The problem is that this pushes the line number off by one and this has a detrimental affect on code that relies on sourcemaps. I'm sure this is as simple as removing the new line character from the appended wrapping function.

germtb:
I changed removed the new line, added a test and changed a few tests that were expecting the new line character. Fixes #1567.


andyberry88:
Changes look good. I think it would be useful to have a `wrappingFunctionShouldNotAddNewLine` test for CommonJS like you have for NamespacedJS. Then it's defined as a specific requirement rather than being tested indirectly by other tests.

dchambers:
Hi @germtb, you needed to change the label back to `Ready for Dev Review` once you'd taken care of any dev-review feedback you got. Doing this now...

PR #1587, opened 2015-11-13
Added a special case to deal with the possibility that the normalizedPath should be an empty string #1487

### LINKED ISSUE #1487: App deployed in tomcat root causes extra slash to appear in url
When deploying a brjs app in tomcat root requests made without a locale get redirected to a url with an extra slash.

E.g: Requesting http://domain.com, will redirect to http://domain.com//en

Requesting a url with a locale specified, no extra slash is added

If you require additional info let me know

germtb:
Fixes #1487


brjs-sdk/sdk/libs/javascript/br-locale/test-unit/tests/br-locale-utility-test.js
```diff
@@ -28,5 +28,9 @@ require('jasmine');
 		it('keeps the query string and anchor in the correct order for urls that contain an html suffix', function() {
 			expect(getLocalizedPageUrl('http://acme.com:1337/app/index.html?query=1#anchor', 'en_GB')).toBe('http://acme.com:1337/app/en_GB.html?query=1#anchor');
 		});
+
+        it('can deal with the special case when the normalized path should be empty', function() {
```
andyberry88: what is 'the special case' here? Would be better named `URL is correct for apps deployed deployed at the root URL`.

andyberry88:
Changes look good. One minor suggestion on the test name.

PR #1589, opened 2015-11-17
Validating and Focus Flags. (to be merged 01/04/16)

ProggerPete:
This patch adds a flag to track whether validation is currently in progress. This is important for fields where the validation is asynchronous. We generally do not want to be displaying an error while this is occurring, but we also want to ensure we have a mechanism to disable other functionality until the field has been validated.

Also included in this patch is a hasFocus attribute, which can be used by controls to indicate that the field is focused. This is useful for date fields, as the field is not validated until it has lost focus. When this is the case we want to disable other functionality until we know that the date is valid.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/node/Field.js
+++ b/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/node/Field.js
@@ -70,6 +70,20 @@ function Field(vValue) {
 	this.visible = new WritableProperty(true);
 
 	/**
+	 * A boolean property representing whether the input field is currently validating
+	 * @type br.presenter.property.WritableProperty
+	 */
+	this.isValidationInProgress = new WritableProperty(false);
+
+	/**
+	 * A boolean property representing whether the input field is currently focused.
+	 * Not all fields will care about tracking whether they have focus or not. For these fields leaving hasFocus
+	 * as undefined seems the most appropriate choice.
+	 * @type br.presenter.property.WritableProperty
+	 */
+	this.hasFocus = new WritableProperty(undefined);
+
+	/**
 	 * The logical control-name the field is being bound to &mdash; this
 	 * value will appear within the <code>name</code> attribute if being bound
 	 * to a native HTML control.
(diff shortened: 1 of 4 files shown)


dchambers:
Hi @ProggerPete, thanks for the commit! This all looks good functionally. The big concern is that `Field` is already too heavy in terms of memory usage and performance, so we'll have to see how it works out in practice in test...

thecapdan:
comparing times it takes to run tests using presenter on IE8, there seems to be a 30% slowdown after your changes @FroggerPete. Those tests are already quite a burden on a few builds so i don't think we can afford that much of a regression

ProggerPete:
That is a pretty gruesome result. I'm amazed given how small the changes are. I feel like either validation must be being called way more often than it should or WritableProperty.setValue is insanely expensive. Possibly both of these. Do the tests give an indication of what is using up all this extra time?

I've suggested this change in response to real problems (UNICRCT4EN-988 among others).

The root issue is that validation isn't always synchronous. As things are currently that means hasError stays at whatever value it had before the change until the asynchronous result arrives. If we're using the hasError field to disable something like a form submit button then this gives the user a window during which they can submit bad data.

One possible work around that we tried was setting hasError to false manually before updating the value. This stops the user from being able to submit data, but results in displaying error messages for data that is error free. Not a great experience.

Being able to track that validation is currently taking place seems essential to me.  My next best option is to add these additions to the caplin.fx.date.DateField class. This should solve my immediate issue, but I think we'll run into the same sorts of problems whenever asynchronous validation is required.

thecapdan:
I did run the tests twice to make sure it wasn't some other anomaly, but you never know. I can have another look today/tomorrow to check the results and maybe do some profiling and find out what's causing slowdown

thecapdan:
Hi @ProggerPete . Having retested your change in more modern browsers, the performance impact is minimal. the reason the impact appeared greater in ie8 may have been since these tests were already skirting close to ie8s stack limit, and it didn't take much to push it over the edge.

I'll discuss the change with the rest of the team and have another look at the impact it may have on our ie8 builds

dchambers:
Hi @thecapdan, where are we regarding this pull-request, I can no longer remember?

thecapdan:
@dchambers while the performance impact is not as bad as we first thought, this change will push ie8 tests over a threshold which cause them to crash, so it will break ie8 builds for other teams. so i think we should wait for a firm decision on IE8 support before we merge this one

dchambers:
Okay, thanks @thecapdan.

thecapdan:
Given that the reason we couldn't merge this was because it would break IE8 builds, since we are dropping support for that browser at the end of march, we can put this on the backlog to be done on April 1st.

dchambers:
Yes, makes sense. Might make sense to add a reminder to your calendar so we don't forget though.

thecapdan:
No need, it's only 10 weeks away, I'll remember

thecapdan:
this work is no long necessary, pending field added here https://github.com/BladeRunnerJS/brjs/issues/1703

PR #1599, opened 2015-11-19
Implemented a check so the RESTful servlet is only usable by allowed IPs #1260

### LINKED ISSUE #1260: Update RESTful servlet to only allow connections from localhost
At the moment it's possible for one developer to interact with the dashboard and dashboard services on another developer's machine. By updating _dashboard-services_ to only accept connections from `localhost` (using [`getRemoteAddr()`](http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getRemoteAddr())), we can easily prevent this from being misused.

germtb:
Fixes #1260


dchambers:
@germtb, is there a reason this pull-request isn't labelled for inclusion into the next release?

PR #1602, opened 2015-11-19
check i18n command

### LINKED ISSUE #1580: Plugin to search for missing i18n tokens
A brjs command that scans an app for all translation tokens (in js, html and xml) and outputs any tokens which are missing for each local.

Example usage:

```
brjs check-i18n myApp

Locale: en
myapp.mybladeset.friendly.message

Locale: de
myapp.mybladeset.myblade.buttonlabel
```

The default command will scan for all locales in the entire app, passing a locale as an arg will just return missing tokens for that locale.

Possible extension:
1. search for unused tokens (so you can find and delete redundant tokens).
2.  being able to specify a subdirectory so you can check i18n coverage for a given blade for example.
3. coverage for tokens that are concatenated in the code e.g. token = "myapp.myblade" + val + "button"

thecapdan:
fixes #1580


andyberry88:
:+1: Nice command @thecapdan

PR #1608, opened 2015-11-24
ensuring ie8 doesnt blow up on i18n warnings when console isnt open

thecapdan:
apps with missing i18n tokens wouldnt open in ie8 due to console warning. (window.console undefined... that old chestnut!)

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-i18n/src/br/i18n/Translator.js
+++ b/brjs-sdk/sdk/libs/javascript/br-i18n/src/br/i18n/Translator.js
@@ -386,6 +386,10 @@ Translator.prototype._getTranslationForKeyOrUndefined = function(token) {
 	token = token.toLowerCase();
 
 	if (!this.tokenExists(token)) {
+		if (typeof window.console === "undefined") {
+			window.console = {};
+			window.console.warn = function(){};
+		}
 		var logConsole = (window.jstestdriver) ? jstestdriver.console : window.console;
 		logConsole.warn('Unable to find a replacement for the i18n key "' + token + '"');
 	}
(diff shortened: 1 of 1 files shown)


brjs-sdk/sdk/libs/javascript/br-i18n/src/br/i18n/Translator.js
```diff
@@ -386,6 +386,10 @@ Translator.prototype._getTranslationForKeyOrUndefined = function(token) {
 	token = token.toLowerCase();
 
 	if (!this.tokenExists(token)) {
+		if (typeof window.console === "undefined") {
```
andyberry88: It would probably be better to conditionally log to the console rather than replacing the console. e.g.
```
var logConsole = (window.jstestdriver) ? jstestdriver.console : window.console;
if (logConsole) {
logConsole.warn('Unable to find a replacement for the i18n key "' + token + '"');
}
```

thecapdan: that won't fix the problem, the problem isn't that console.warn is undefined, its that window.console is undefined in ie8 when the console isnt open

andyberry88: Ahh yes ok, misread your comment. My first code snippet (before the 2 edits) is correct then. Something like:

```
var logConsole = (window.jstestdriver) ? jstestdriver.console : window.console;
if (logConsole && logConsole.warn) {
    logConsole.warn('Unable to find a replacement for the i18n key "' + token + '"');
}
```

thecapdan: perfect :+1:

PR #1614, opened 2015-12-02
Bugfix clearnode now removes the node

thecapdan:
previously the remove method was being passed a string which would not clear the node

### DIFF
--- a/brjs-core/src/main/java/org/bladerunnerjs/model/engine/AbstractRootNode.java
+++ b/brjs-core/src/main/java/org/bladerunnerjs/model/engine/AbstractRootNode.java
@@ -89,7 +89,7 @@ public abstract class AbstractRootNode extends AbstractNode implements RootNode
 	public void clearRegisteredNode(Node node) {
 		String normalizedPath = node.dir().getAbsolutePath();
 		List<Node> nodesForPath = nodeCache.get(normalizedPath);
-		nodesForPath.remove(node.getTypeName());
+		nodesForPath.remove(node);
 	}
 	
 	@Override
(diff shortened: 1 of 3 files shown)


brjs-core/src/main/java/org/bladerunnerjs/model/engine/AbstractRootNode.java
```diff
@@ -89,7 +89,7 @@ public void registerNode(Node node) throws NodeAlreadyRegisteredException
 	public void clearRegisteredNode(Node node) {
 		String normalizedPath = node.dir().getAbsolutePath();
 		List<Node> nodesForPath = nodeCache.get(normalizedPath);
-		nodesForPath.remove(node.getTypeName());
+		nodesForPath.remove(node);
```
andyberry88: Is there a test that proves this should be `node` not `node.getTypeName()`?

dchambers: Is that a rhetorical question :smiley:

andyberry88: No. Seems to be an arbitrary change where nothing broke before and nothing breaks now. So why do we need the change? Or is that line even needed if it never worked....?

thecapdan: That method is not covered by a test + i can't find anywhere where its being used, so maybe it can be removed? But, List remove method takes an object, not a string, so applying the fix won't do any harm, so we should either apply the fix, or delete the method

andyberry88: There's no point applying a fix to a piece of code that isn't even used :-) Can we try removing it? `nodeCache.get(..)` could return a copy of the array it holds, in which case that method would have no effect on the real state anyway.

PR #1622, opened 2015-12-07
Add `service!$data` source module

andyberry88:
Add `service!$data` source module that lists services and their dependencies.

There *could* be a slight performance/memory hit here since in order to detect services in NamespacedJS we have to do a second pass looking for `getService('...')` calls. I've done some basic investigation and it seems to be negligible, but we should probably double check this with a larger app.


thecapdan:
performance impact is not detectable in our test-app (fxtrader).

require('service!$data') behaves as expected. test coverage looks good

Issue #1637, opened 2016-01-14
improve ErrorMonitor with its own hasError property


mborik:
Improve ErrorMonitor to has its own hasError property which indicates if any of the associated fields has error. Also fixed **_addTooltipTo** private method to show tooltips only if it has failureMessage.

PR #1638, opened 2016-01-14
improve ErrorMonitor with its own hasError property #1637

mborik:
Improved ErrorMonitor with its own hasError property which indicates if any of the associated fields has error.
Fixed **_addTooltipTo** private method to show tooltips only if it has failureMessage.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/util/ErrorMonitor.js
+++ b/brjs-sdk/sdk/libs/javascript/br-presenter/src/br/presenter/util/ErrorMonitor.js
@@ -2,6 +2,7 @@
 
 var ToolTipField = require('br/presenter/node/ToolTipField');
 var PropertyHelper = require('br/presenter/property/PropertyHelper');
+var EditableProperty = require('br/presenter/property/EditableProperty');
 var Errors = require('br/Errors');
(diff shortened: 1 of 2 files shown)


thecapdan:
Hi @mborik 

This has caused some failures in our presenter ATs.

I believe the issue is with _addError is now being called too early. Where you have added this:

```
if (this.hasError.getValue() === true) {
	oTicketErrorProperty._addError(oField);
}
```

It is trying to add the error before the field has been updated. If i comment out oTicketErrorProperty._addError(oField), everything passes (it waits for the observable instead)

Would you mind having a look.

Thanks

mborik:
Hi @thecapdan,
thanks for your comment. I dealt with the issue this morning and I hope that I finally solved it. As you mentioned, I think that issue is in ``fValidationSuccessHandler`` function condition: there is not enough to test only ``hasError`` of the field to true but also there is should be test if field has ``failureMessage`` non-empty.
After that, ATs passing without any problem and also behavior in the application is correct.
I'll push my changes afternoon.

thecapdan:
great. thanks

thecapdan:
hi @mborik , would you mind merging master into your branch, I can then merge this and move it to done. thanks

PR #1639, opened 2016-01-21
Autocomplete improvements

dogoku:
This is a PR consisting of a number of improvements/changes for the JQueryAutoCompleteControl

I've split the changes into seperate commits, to make things a bit easier. Some commit messages have more than one lines so do bear that in mind.


dchambers:
Hi @dogoku, thanks for the pull-request :smile:.

Before we can proceed there is one commit that will need to be reverted, and I noticed that there are no tests for any of these changes. It looks like there are tests in `JQueryAutoCompleteControlAdapterTest` that are relevant to this, or are these not for the same thing?

dogoku:
Hello, I've updated my commits and added tests in all of them

Unfortunately, github doesn't handle rebases/squishes as well as Stash does, so you can't really notice something happened, unless you actually go into the particular commit and know it's previous state.

Regardless, I beliebe this PR is good to go

PR #1654, opened 2016-02-03
fix: Change br-bootstrap exports to null

briandipalma:
Exporting an object literal caused issues for the brjs-app-converter.


briandipalma:
Minor change, the conversion tool puts the export on the `window` unless it's `null` since nothing is exported from this library anyway I thought we may as well just change the manifest.

dchambers:
Dev reviewed :+1:

PR #1656, opened 2016-02-05
refactor: Inline app-metadata require

briandipalma:
The app-metadata module ID has been renamed so it doesn't clash with the module ID for aliases `$data`.

I haven't run the build locally as it's a bit awkward to have to install multiple Java's but I'm sure this will fail. I don't really expect this to be merged, it's more to give an idea as to what JS changes are required for CTNXT-21. 

cc: @bit-shifter as I imagine you will be picking up that JIRA?


germtb:
Reviewed

bit-shifter:
Looks good to me - @Aivarass can you pick this up?

Issue #1653, opened 2016-02-02
Locale switcher returns incorrect localised url in IE


thecapdan:
Failing test in br-locale-utility-test.js:

```
 switcher.getLocalizedPageUrl(). appends the locale to the end of page urls that have no folder
         failed (1.00 ms): Error: Expected 'http://acme.com:80//en_GB' to be 'http://acme.com/en_GB'.
```

PR #1661, opened 2016-02-10
Add new test and fix different behaviour in ie

germtb:
Fixes https://github.com/BladeRunnerJS/brjs/issues/1653


Aivarass:
This fix works for IE8 / IE9 / IE10
However tests still fails for IE11, because it has a different user agent app name, it's called "Netscape". This is very confusing because Chrome and Firefox are also called Netscape. I think that playing around with regex while using navigator.userAgent output might solve it. In addition, not sure what is the behavior of Microsoft Edge (which is also called Netscape), will try to test it and update.

PR #1668, opened 2016-02-23
Enable node style requires for index js files

bit-shifter:
Part of the brjs-ify work (converting node style libs -> brjs conformant libs), adds facility to require the index js file from within a brjs conformant lib.


dchambers:
Dev reviewed :+1:

Issue #1681, opened 2016-02-29
Angular2 basic app/blade, and library support


oniseijin:
Basic support within template and library for Angular2. 

Requires: 
-Libraries, templates

Long story short, I have already gotten a basic version working with angular umd. I'll make a branch and point to it when there is a more complete set of templates/libraries, but I wanted to point out early on to others that I have indeed proved it possible in BRJS.

Needed umd versions, es5-shim, es6-promise, rx. Angular2 feels a bit more straightforward than angular1, and I think it would be good to get going from the start, although this may be "too" early. 

Interested in feedback as this goes a bit further. For me, there is enough basic plumbing in place to keep pursuing. 

It took a lot of research to find a way in ES5 to do this, especially the class level annotation array, and most samples are ES6 or TypeScript: in short, we are back to not having to deal with controller functions and $scope
<pre>
'use strict';

var i18n = require('br/I18n');
var ng = require("angular2");

function @bladeTitleComponent() {

	this._eventHub = require('service!br.event-hub');
	this.welcomeMessage = 'Welcome to your new Blade.';
	this.buttonClickMessage = i18n('@bladeNamespace.button.click.message');
	this._logWelcome();

}

// has to be class level
@bladeTitleComponent.annotations = [
 new ng.core.Component({
	 selector: '@bladeTitle' // adjust as needed
 }),
 new ng.core.View({
	 template: require('service!br.html-service').getTemplateElement( '@bladeNamespace.view-template' ).outerHTML
 }),

];

@bladeTitleComponent.prototype._logWelcome = function() {
	console.log(this.welcomeMessage);
};

@bladeTitleComponent.prototype.buttonClicked = function() {
	console.log('button clicked');
	var channel = this._eventHub.channel('@blade-channel');
	channel.trigger('hello-event', {some: 'Hello World!'});
};

module.exports = @bladeTitleComponent;

</pre>

Template is: 

```
<div id="@bladeNamespace.view-template">
	<div class="@bladeset-@blade-blade">
		<h1 class="hello-world-message">
			 @{@bladeNamespace.hello.world}
		</h1>
		<p class="welcome-message">{{welcomeMessage}}</p>
		<p>Try using alternative stylesheets to switch the theme.</p>
		<p>{{buttonClickMessage}}</p>
		<button class="button" (click)="buttonClicked()">Say Hello!</button>
	</div>
</div>
```

Workbench bootstrap: 
```

[...]

PR #1687, opened 2016-03-09
angular2 templates and sdk support #1681 

oniseijin:
Initial pull request for #1681. 

Aspect, blade, tests, copied over App, Bladeset defaults. Using ES5 structure, but needed a few libs. Fully functional compared to Knockout default. 

However, acceptance test doesn't use angular-test/element checks, as there is an issue with the version of jasmine, so that would be later.


thecapdan:
this can be closed, these changes were merged here https://github.com/BladeRunnerJS/brjs/pull/1697

PR #1688, opened 2016-03-10
feat: Allow i18n messages to be registered with i18n Translator

briandipalma:
cc @bit-shifter @thecapdan 

The webpack caplin i18n loader registers i18n tokens on load up so we need to provide a mechanism in the Translator class to support this approach.

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-i18n/test-unit/tests/TranslatorTest.js
+++ b/brjs-sdk/sdk/libs/javascript/br-i18n/test-unit/tests/TranslatorTest.js
@@ -9,7 +9,7 @@
 	TranslatorTest.prototype.setUp = function() {
 		JsHamcrest.Integration.JsTestDriver();
 		JsMockito.Integration.JsTestDriver();
-		
+
 		this.subrealm = realm.subrealm();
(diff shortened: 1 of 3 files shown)


brjs-sdk/sdk/libs/javascript/br-i18n/test-unit/tests/TranslatorTest.js
```diff
@@ -411,4 +411,18 @@
...
+			upperCaseKey: 'Value'
+		};
+		var Translator = require('br/i18n/Translator');
+		var translator = new Translator(this.i18nTimeDateNumberMessages, 'locale');
+
+		translator.registerTranslations('locale', translations);
+
+		assertEquals(translator.getMessage('key'), 'value');
+		assertEquals(translator.getMessage('upperCaseKey'), 'Value');
+	};
```
thecapdan: do we need a test that ensures this method doesn't override existing translations?

briandipalma: Good idea, I've added a test like that.

PR #1689, opened 2016-03-15
Add debug option to brjs serve

germtb:
This change in the brjs script allows the option to debug the java code of brjs by using `brjs debug serve`, which is can be useful at times.

I have made it work with intelliJ on windows using [this](https://www.linkedin.com/pulse/debug-jar-files-intellij-idea-maksym-lushpenko) tutorial. It might need testing on other OS.


thecapdan:
this works for bash and cygwin. @germtb can we make this work for cmd?

PR #1691, opened 2016-03-21
fix: Register a translation without predeclaring its locale

briandipalma:
Previously the user would need to predeclare the locales that would be
registered (for example in the index.html page). This is annoying when
testing code that requires translations or when testing in node.

closes: CTNXT-26


briandipalma:
Failure seems to be that it can't download gradle...

PR #1692, opened 2016-03-22
Add mingw64 support

germtb:
Update the brjs.sh so that it work with 64 bits git bash versions.


thecapdan:
to be reopened without the debug changes

PR #1695, opened 2016-04-01
feat(i18n): If a locale isn't specified use en as the default

briandipalma:
This means UTs in the CT.Next stack can skip the boilerplate of setting the locale.


brjs-sdk/sdk/libs/javascript/br-i18n/src/br/i18n/Translator.js
```diff
@@ -46,7 +46,7 @@ function Translator(messageDefinitions, useLocale) {
 
 	/** @private */
 	this.localizationPrefs = {};
-	this.setLocale(useLocale);
+	this.setLocale(useLocale || 'en');
```
thecapdan: may this cause issues if 'en' is not one of the defined locales for an app, if so couldn't we use 'defaulLocale' which is defined above? or is this for the case when no locales have been defined?

thecapdan: @briandipalma also, could you merge master into this branch. ty

briandipalma: If there is no locale defined in  `useLocale` and `en` doesn't exist then there might be issues but this shouldn't be possible in a BRJS app, I believe it expects a locale to be defined?

From my reading of the code `defaultLocale` is badly named. It is in fact a `productionFallbackLocale` as it's a fallback that only works in production

```javascript
var message = this.messages[token];
if (typeof message === 'undefined') {
	log.warn(Translator.MESSAGES.UNTRANSLATED_TOKEN_LOG_MSG, token, this.locale);
	if (!require('service!br.app-meta-service').isDev()) {
		message = this.defaultLocaleMessages[token];
	}
}
```

If we were to use the fallback the UTs will still fail as it only works in prod, the `Translator` would return the i18n token key and not the translated value which is what the tests in mobile expect. As you say yourself this change is for when no locales have been defined, in prod you would expect a locale to be defined.

PR #1696, opened 2016-04-04
Add support for git bash 64 bits version

germtb:


thecapdan:
there is probably a simpler way to write the conditional, but it is clear the way it is now so we can leave it as it is. done

PR #1706, opened 2016-05-17
fix: Remove dynamic require from AliasRegistryClass

briandipalma:
Allows us to prevent the webpack bundling errors around `AliasRegistryClass`.

closes: CTNXT-193


bit-shifter:
This allows us to bypass the erroneous inclusion of addition source and resource files from base directories when using webpack.

see http://webpack.github.io/docs/context.html

PR #1708, opened 2016-06-07
Protect from an exception on writing to readonly property

vmus:
KeyboardEvent.key has been implemented in Chrome 51+ and it is readonly
http://caniuse.com/#feat=keyboardevent-key
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility
https://bugs.chromium.org/p/chromium/issues/detail?id=227231

Attempts to modify this property now result in an exception:
`TypeError: Cannot assign to read only property 'key' of object '#<KeyboardEvent>'`


thecapdan:
closing in favour of https://github.com/BladeRunnerJS/brjs/pull/1711

PR #1713, opened 2016-06-23
Fix datepicker i18n typo

vmus:


bit-shifter:
Good catch, thanks for the PR.

PR #1717, opened 2016-07-19
fix(br-validation): Separate require from constuction of `ValidationR…

briandipalma:
…esult`

Inlining the two causes webpack to complain about not being able to statically analyze
the require.

closes: CTNXT-278


bit-shifter:
all good, no-op change

PR #1722, opened 2016-09-01
task(i18n): Separate i18n token storage from Translator

briandipalma:
This prevents circular dependency issues where `Translator` requires i18n tokens
but that in turn requires an instanciated `Translator`.

closes: CTNXT-470


brjs-sdk/sdk/libs/javascript/br-i18n/src/br/i18n/I18nStore.js
```diff
@@ -0,0 +1,64 @@
...
+
+	get defaultMessages() {
+		if (this.messageDefinitions[this.defaultLocale] === undefined) {
+			this.messageDefinitions[this.defaultLocale] = {};
+		}
+
+		return this.messageDefinitions[this.defaultLocale];
+	},
+
+	get messages() {
```
thecapdan: @briandipalma when running this with closure-medium minifier we get the following error:

ERROR - Parse error. getters are not supported in older versions of JavaScript. If you are targeting newer versions of JavaScript, set the appropriate language_in option.
	get messages() {

This happens in tests and when running the it-app with js.bundle dev-minifier="closure-medium"

trueadm:
Reviewed

Issue #1728, opened 2016-09-05
Remove redundant shims and shivs


thecapdan:
now that we no longer support IE8 there are many shims that can be removed (some would potentially cause issue in webpack)

PR #1729, opened 2016-09-05
remove redundant shims and shivs not that we no longer support IE8 fi…

thecapdan:
…xes #1728


thecapdan:
reviewed by @briandipalma

PR #1734, opened 2016-09-07
replace one use of Number.isNaN with es5 equivalent to ensure it can …

### LINKED ISSUE #1733: Presenter uses Number.isNaN function which is not supported by IE11
We previously had an IE8 shim which covered this. We need to find an alternative

thecapdan:
…work in IE11. fixes #1733


briandipalma:
Looks good to me.

PR #1744, opened 2016-11-08
changing ListenerCompatUtil to use typeof when checking listener meth…

### LINKED ISSUE #1743: ListenerCompatUtil incorrectly evaluates method
...  evaluation is taking place in a popout and it is not identified as an ```instanceof Function``` but ```typeOf method``` returns "function".

thecapdan:
…od is a function. fixes #1743


thecapdan:
pretested on downstream projects

PR #1745, opened 2016-11-24
bugfix for locale switcher, ensures locale redirect works form root l…

thecapdan:
…evel index.html.

- contains check for IE
- unit test added
- release note added


thecapdan:
manually tested by mr south

PR #1756, opened 2016-12-12
1755/bugfix ensure svg elements work in ie11

### LINKED ISSUE #1755: Cover edge case with SVG Elements in IE11
When using IE11, SVG elements can't use 'classList'. Used in br/util/ElementUtility.

This class could also do with some additional unit tests

thecapdan:
fixes #1755


briandipalma:
Looks good to me

PR #1758, opened 2017-02-08
#1757 ensure fixture factory teardowns are called during GWTTestRunne…

### LINKED ISSUE #1757: GWTTestRunner should call fixture factory teardown
start test calls setUp method
end test should call tearDown

This will allow users to clear up services and mappings generated in the fixture factory

thecapdan:
…r endTest

fixes #1757

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br-test/src/br/test/GwtTestRunner.js
+++ b/brjs-sdk/sdk/libs/javascript/br-test/src/br/test/GwtTestRunner.js
@@ -243,6 +243,15 @@ GwtTestRunner.prototype.endTest = function() {
 
 	}
 
+	if (this.m_oFixtureFactory.tearDown) {
+		try {
+			this.m_oFixtureFactory.tearDown();
+		} catch (e) {
+			throw new Errors.CustomError("TestTearDownError", e.message,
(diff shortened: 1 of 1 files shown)


brjs-sdk/sdk/libs/javascript/br-test/src/br/test/GwtTestRunner.js
```diff
@@ -243,6 +243,15 @@ GwtTestRunner.prototype.endTest = function() {
 
 	}
 
+	if (this.m_oFixtureFactory.tearDown) {
+		try {
+			this.m_oFixtureFactory.tearDown();
+		} catch (e) {
+			throw new Errors.CustomError("TestTearDownError", e.message,
```
briandipalma: I don't know, I'm not sure I get the point of catching an error and then throwing your own. Why not just let the original error bubble up?

thecapdan: This is how the rest of the errors are handled in GWTTestRunner, I tried to keep it consistent - not sure why it was originally done like this, could be to avoid jstestdriver garbling errors, which its notorious for doing

briandipalma: Cool, fair enough.

PR #1760, opened 2017-02-16
feat(aliases): Simplify the alias data format and allow alias registr…

briandipalma:
…ation

closes: CTNXT-608, CTNXT-583

### DIFF
--- a/brjs-sdk/sdk/libs/javascript/br/src/br/AliasRegistryClass.js
+++ b/brjs-sdk/sdk/libs/javascript/br/src/br/AliasRegistryClass.js
@@ -36,10 +36,9 @@ var dynamicRefRequire = require('br/dynamicRefRequire');
 * discover all of the classes which implement a particular interface, which
 * makes it a good candidate for creating SPI type, auto-discovery mechanisms.</p>
 */
-var AliasRegistryClass = function(aliasData)
-{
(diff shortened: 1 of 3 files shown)


briandipalma:
For some reason the build won't run on my machine, pushing this to trigger a build and see the output.

brjs-sdk/sdk/libs/javascript/br/src/br/AliasRegistryClass.js
```diff
@@ -128,13 +127,24 @@ AliasRegistryClass.prototype.isAlias = function isAlias(aliasName) {
...
+AliasRegistryClass.prototype.registerAliasProvider = function(aliasName, aliasProvider) {
+	this._aliasData[aliasName] = aliasProvider;
 };
 
 /**
  * @private
  */
 AliasRegistryClass.prototype._getClassRef = function(aliasName) {
+	// Do you access the alias via a provider function.
+	if (typeof this._aliasData[aliasName] === 'function') {
```
bit-shifter: With this change, when a call is made to `AliasRegistry#getClass(alias)` from `ServiceRegistry`, we will no longer verify that the class referenced by the alias adheres to any implemented interfaces (as there is no longer an interface property to verify against in alias data).

PR #1787, opened 2017-06-26
Update Selenium to version 3.4.0

vmus:


vmus:
Needed the latest geckodriver

0.16.0 (2017-04-21)
- [`SendKeysParameters`](https://docs.rs/webdriver/0.25.0/webdriver/command/struct.SendKeysParameters.html), which is used for the [Element Send Keys](https://w3c.github.io/webdriver/webdriver-spec.html#element-send-keys) and [Send Alert Text commands](https://w3c.github.io/webdriver/webdriver-spec.html#send-alert-text), has been updated to take a string `text` field (https://github.com/mozilla/geckodriver/issues/659)

However, there are still open issues that may be important to someone else:
- https://github.com/mozilla/geckodriver/issues/786
- https://github.com/mozilla/geckodriver/issues/646

PR #1789, opened 2017-07-12
Pctlibrary 1935/switching to typeof in observable class

thecapdan:


brjs-sdk/sdk/libs/javascript/br-util/src/br/util/Observable.js
```diff
@@ -67,11 +67,11 @@ Observable.prototype.getCount = function() {
...
 Observable.prototype.addObserver = function(observer) {
 	if (
-		!(observer instanceof Object) ||
-		(observer instanceof String ||
-			observer instanceof Number ||
-			observer instanceof Boolean ||
-			observer instanceof Function)
+		!(typeof observer === "object" && observer !== null) ||
+		(typeof observer === "string" ||
+			typeof observer === "number" ||
```
bit-shifter: NaN would pass the number check here - changing this to:

`(typeof observer === "number" && !isNan(observer))`

should mitigate this.

I notice that the tests using new (Number|String|Boolean) are failing due to their typeof check resolving to  `object`.

Whilst this change is a compatibility breaking change, I'm inclined to remove those tests, as it's an uncommon form (and avoided by most developers).

We will need to add a release note to mention that the form is not supported.

